PR #6, opened 2018-06-28
Materials to "Generating Random Data in Python"

bsolomon1124:


dbader:
@bsolomon1124 Setting up the article on RP now, LGTM :)

PR #8, opened 2018-07-30
Correction tag name capitalization

aln787:
Upper and lower case [p]eople show up as different endpoint groups in the swagger ui.


dbader:
Thanks for the PR @aln787 😄 Pinging @writeson here because he wrote the original article and coded up the sample project.

PR #13, opened 2018-09-07
Materials for opencv color spaces article

dbader:
@ysbecca Opened a PR for your branch so the linter runs :)


dbader:
@ysbecca Looks like there are a couple of linter warnings to resolve but other than that this looks good!

ysbecca:
Hey @dbader, I read somewhere that we don't mind the "line too long" errors. Please let me know if that is incorrect; I can shorten the lines. Also, Axes3D is imported and actually needed, even if not explicitly used. I'm not sure how to deal with that issue.

dbader:
@ysbecca Ah yeah, the "don't wrap lines" rule only applies to Markdown. For Python code we're using the flake8 linter with standard settings, which means a 79 character line length limit as per PEP 8 :)

ysbecca:
@dbader As far as I am aware, the way matplotlib 3D axes works is that you have to import it. That allows for using the parameter projection="3d". I have seen no other way to do it... unfortunately, that causes the linter to fail as it thinks the import is being unused, when actually it is. 

Please let me know what you suggest! All the other errors are resolved now.

opencv-color-spaces/finding-nemo.py
```diff
@@ -0,0 +1,166 @@
...
+Author: Rebecca Stone @ysbecca
+'''
+
+import cv2
+import numpy as np
+
+import matplotlib.pyplot as plt
+from matplotlib import colors
+
+from mpl_toolkits.mplot3d import Axes3D  # noqa
```
dbader: I added a "noqa" comment which disables flake8 on that particular line.

dbader:
@ysbecca Perfect, thanks. LGTM 😄

PR #24, opened 2019-01-05
Add materials for type checking article

gahjelle:
Code examples from https://realpython.com/python-type-checking/


dbader:
LGTM :-) 👍

PR #28, opened 2019-01-21
Flask connexion rest part 3

writeson:


dbader:
Quick heads up @writeson this still triggers a couple of preflight checks that need resolving:

https://realpython.com/preflight/flask-connexion-rest-api-part-3/

The other thing we need to do is move the example code to the https://github.com/realpython/materials repo in a separate PR (so we don't have to maintain it in two separate places).

dbader:
Maybe I'm blind but it looks like this PR is missing the actual article text? 😃

dbader:
I'm going to merge this @writeson so your article links start working. If you need to make updates—no problem, just create another PR please :)

PR #34, opened 2019-02-06
Added example project for article 'Django Migrations: A Primer

dhepper:


dhepper:
@dbader this PR fails the checks because includes non-compliant code generated by Django. I'm reluctant to change the code because then it is not what readers get when they follow along with the article. How would you deal with that?

dbader:
> @dbader this PR fails the checks because includes non-compliant code generated by Django. I'm reluctant to change the code because then it is not what readers get when they follow along with the article. How would you deal with that?

Hmm good point, I'm guessing this is the problematic line?

```
./django-migrations/bitcoin_tracker/historical_data/migrations/0001_initial.py:17:80: E501 line too long (114 > 79 characters)
```

I think we should just run `black` on the whole thing to reformat it and it'll be alright. It's not 100% what readers will get in the autogenerated code—but because it's autogenerated it'll likely change anyway with future releases of Django etc.

dhepper:
Ok, I'll reformat with black then and add a remark to the readme. But what about the unused imports? These are also created by Django. Possible solutions:

* remove the files
* remove the imports
* add `#noqa` to the unused imports

> ./django-migrations/bitcoin_tracker/historical_data/models.py:4:1: E302 expected 2 blank lines, found 1
> ./django-migrations/bitcoin_tracker/historical_data/admin.py:1:1: F401 'django.contrib.admin' imported but unused
> ./django-migrations/bitcoin_tracker/historical_data/tests.py:1:1: F401 'django.test.TestCase' imported but unused
> ./django-migrations/bitcoin_tracker/historical_data/views.py:1:1: F401 'django.shortcuts.render' imported but unused
> ./django-migrations/bitcoin_tracker/historical_data/migrations/0001_initial.py:17:80: E501 line too long (114 > 79 characters)
> ./django-migrations/bitcoin_tracker/bitcoin_tracker/settings.py:90:80: E501 line too long (91 > 79 characters)
> ./django-migrations/bitcoin_tracker/bitcoin_tracker/settings.py:93:80: E501 line too long (81 > 79 characters)
> ./django-migrations/bitcoin_tracker/bitcoin_tracker/settings.py:96:80: E501 line too long (82 > 79 characters)
> ./django-migrations/bitcoin_tracker/bitcoin_tracker/settings.py:99:80: E501 line too long (83 > 79 characters)

dbader:
Thanks!

> * add `#noqa` to the unused imports

@dhepper That would be my preferred solution

PR #36, opened 2019-02-24
Jima/threading

jima80525:
Source code for upcoming article on threading.


dbader:
@jima80525 This looks good! One small request: Let's change the folder name from `_` underscores to `-` dashes to be consistent with the rest of the repo. After that, LGTM! :-)

(This also means we'll need to update the link in your upcoming article.)

PR #62, opened 2019-08-08
Update views.py

udayposia:
Using **get** the method we will not get iterable value, rather use **filter** to get the proper result.


DmytroLitvinov:
I think it is better to use even here [get_object_or_404](https://docs.djangoproject.com/en/2.2/topics/http/shortcuts/#get-object-or-404) for `Post` model.

udayposia:
@DmytroLitvinov , using get is raising an error that " value is not iterable. "

PR #74, opened 2019-08-15
Simpy

jayascript:
**Where to put new files:**

4. 🎉

### DIFF
--- a/simulation-with-simpy/simulate.py
+++ b/simulation-with-simpy/simulate.py
@@ -9,6 +9,9 @@ SimPy version: 3.0.11
 import simpy
 import random
 
+arrival_times = []
+departure_times = []
+
 
 class Theater(object):
     def __init__(self, env, num_cashiers, num_servers, num_ushers):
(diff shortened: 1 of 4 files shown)


simulation-with-simpy/simulate.py
```diff
@@ -0,0 +1,118 @@
...
+    with theater.cashier.request() as request:
+        yield request
+        yield env.process(theater.purchase_ticket(moviegoer))
+
+    with theater.usher.request() as request:
+        yield request
+        yield env.process(theater.check_ticket(moviegoer))
+
+    get_food = [0, 1]
+    if random.choice(get_food) == 1:
```
jima80525: This is probably cleaner as:

```
if random.choice([True, False]):
```

simulation-with-simpy/simulate.py
```diff
@@ -0,0 +1,118 @@
...
+        arrival_time = env.now
+        arrival_times.append(arrival_time)
+
+
+def calculate_wait_time(arrival_times, departure_times):
+    total_wait = []
+    total_seated = len(departure_times)
+
+    for i in range(total_seated):
+        total_wait.append(departure_times[i] - arrival_times[i])
```
jima80525: Two different ideas here:

1. since both arrival and departure times are globals, why not replace `departure_times` with `wait_times` and compute each wait time when you know the departure?
2.  If that's not appealing, then this could potentially be a list comprehension using enumerate to get the index from one.

I'd vote for option 1 as it's cleaner. :)

simulation-with-simpy/simulate.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Pretty print the results
+    average_wait = sum(total_wait) / total_seated
```
jima80525: could use `statistics.mean()` here. :)

simulation-with-simpy/simulate.py
jayascript: Done! In addition, I found a way to have `wait_times` as the only global. Check [this comment](https://github.com/realpython/tutorial-drafts/pull/337#discussion_r315884543) on the tutorial PR for more info.

simulation-with-simpy/simulate.py
jayascript: Done!

PR #82, opened 2019-10-02
Updating code to use Geir's codetiming module rather than my homegrow…

writeson:
…n elapsedtime module, which seemed to be causing some confusion.

**Where to put new files:**

4. 🎉


understanding-asynchronous-programming/example_3.py
```diff
@@ -1,15 +1,17 @@
 import time
 import queue
-from lib.elapsed_time import ET
+from codetiming import Timer
 
 
 def task(name, queue):
+    text = ''.join([f'Task {name} elapsed time: ', '{:.2f}'])
+    timer = Timer(text=text)
```
gahjelle: Ah, I see you ran into the same f-string-and-`.format` problem that had me stumped for a while :blush: 

You can escape curly braces in f-strings by doubling them, so `f"{{:.2f}}"` evaluates as `"{:.2f}"`. The instantiation of the timer can then be done with:

```
timer = Timer(text=f"Task {name} elapsed time: {{:.2f}}")
```

gahjelle: I also note that in your current `ET` code you print out with one decimal, while here you've changed to 2. At least on my computer, I then get results like `Task One elapsed time: 15.01` which doesn't look quite as clean and might distract a little from the focus of the article?

I would therefore suggest going back to one decimal by using `.1f` instead of `.2f`.

gahjelle: Finally, I noticed the linter was failing on your updates. This seemed to be solely due to `black` wanting to have double quotes instead of single quotes. I've used double quotes in my suggested code above.

understanding-asynchronous-programming/example_3.py
```diff
@@ -27,18 +29,16 @@ def main():
...
-
-    print(f"\nTotal elapsed time: {et():.1f}")
+    with Timer(text='\nTotal elapsed time: {:.2f}'):
+        while not done:
+            for t in tasks:
+                try:
+                    next(t)
+                except StopIteration:
+                    tasks.remove(t)
+                if len(tasks) == 0:
```
gahjelle: I see this comes from the original code, so I'll let you be the judge if it's worth changing:

I believe it's more "pythonic" to test directly on the falsyness of the list:

```
if not tasks:
    done = True
```

gahjelle: An even bigger change (which would probably force you to update some of the explaining text as well) could be to refactor out the `done` variable completely:

```
with Timer(text="\nTotal elapsed time: {:.1f}"):
    while True:
        for t in tasks:
            try:
                next(t)
            except StopIteration:
                tasks.remote(t)
            if not tasks:
                break
```

Personally, I find this cleaner than using the extra variable, but I'm not sure it's as obvious for everybody? Again, I'll leave it for you to decide.

understanding-asynchronous-programming/example_3.py
writeson: Changed all three to your recommendations, though I'm having trouble getting black to run. Keeps giving me a 

```
dyld: Library not loaded: @executable_path/../.Python
  Referenced from: /usr/local/Cellar/black/19.3b0/libexec/bin/python3.7
  Reason: image not found
```
error. But I changed to double-quotes just with a search and replace as there were only a few.

understanding-asynchronous-programming/example_3.py
writeson: I'm going to leave this as it is, though I agree with your choices. Since the article is already out there this way, I'm going to leave it be for the moment. Next time... :)

PR #87, opened 2019-11-09
Use consistent tag name ('people') across Version 2, 3 and 4

nakami:
Version 3: Change tag of /people endpoint from 'People' to 'people'. This way, there is just one tag ('people') and all endpoints are bundled together in the Swagger UI.

**EDIT**: changed all occurrences of the tag 'People' to 'people' across all versions containing the swagger.yml file.

### DIFF
--- a/flask-connexion-rest/version_2/swagger.yml
+++ b/flask-connexion-rest/version_2/swagger.yml
@@ -16,7 +16,7 @@ paths:
     get:
       operationId: people.read
       tags:
-        - People
+        - people
       summary: The people data structure supported by the server application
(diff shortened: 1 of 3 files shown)


nakami:
There is an inconsistency between the swagger.yml versions of 3 and 4.
In Version 3, only one tag is named 'People', rest uses tag 'people'.
In Version 4, all tags are 'People' (capital P).

nakami:
I realised that part 2, 3 and 4 of the  flask-connexion-rest article use the tag 'People'. Might instead use that tag across every swagger.yml file...

PR #102, opened 2020-03-02
Code to support Arcade Platformer article

JFincher42:
**Where to put new files:**

4. 🎉


gahjelle:
@JFincher42 

Here are some suggested updates to the code. As agreed I have posted them as commits on the branch. I've tried to keep each commit small and focused so that you can review them independently :blush: 

A few comments on the updates:

- Flake gets quite angry at the `*` imports, so I namespaced all use of `constants`. As a test, I ended up renaming `constants` to `game` during import because I thought `game.PLAYER_SPEED` seemed quite intuitive. However, I recognize it might be more readable to simply do `import constants` and then use `constants.PLAYER_SPEED` and so on. Feel free to change that!
- As we talked about, I changed `os.path` to `pathlib`. I think `pathlib` makes all the path joining look much nicer. However, I did realize that `arcade` doesn't quite support `pathlib` so there were a few places the paths needed to be explicitly cast to strings.
- I added type hints on all methods. To truly do static type checking, we should add type hints also for most attributes (especially those initialized to `None` which therefore can't be inferred automatically). However, I'm not sure that is necessary here? After adding type hints, I removed the types from the docstrings as that doesn't then add any info - and potentially becomes easily outdated if we make changes
- I added in explicit use of keyword arguments for most calls to `arcade` function to make it clearer what each argument refers to
- It seems that the issue with the screen not properly updating at the beginning of new levels is due to the logic around whether the viewport should update or not, inside `.scroll_viewport()`. I ended up removing the check and just always update the viewport. This seems to have the start of each level update properly. I also didn't notice any performance hits.
- I also reset the player speed when setting up a new level, as that seemed to be the culprit having the player start running on new levels.
- I did a minor rewrite in `.create_enemy_sprites()`, having it return a list instead of setting the attribute - similarly to how `.create_player_sprite()` works. It also returns an empty sprite list when not on level 2, so we don't need all the tests on whether `.enemies_list` is `None`.
- I tweaked and added some docstrings to make them more or less consistent.

Feel free to skip any of these commits if you disagree with them. And let me know if you have any questions.

bzaczynski:
Looks good to me :+1: We might need to reformat the code with `black` if the linters prevent the branch from merging. Anyway, I'd rather not do that if possible to avoid inconsistency with the article.

jaschmitt (approved):

PR #103, opened 2020-03-08
Python eval mathrepl

lpozo:
**Where to put new files:**

4. 🎉


python-eval-mathrepl/mathrepl.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
```
bryanwweber: I left the same comment on the article, but I don't think you need this line.

python-eval-mathrepl/mathrepl.py
```diff
@@ -0,0 +1,78 @@
...
+
+# File: mathrepl.py
```
bryanwweber: Also probably don't need this line, since the file is named `mathrepl.py` :smile:

python-eval-mathrepl/mathrepl.py
```diff
@@ -0,0 +1,78 @@
...
+
+"""MathREPL, a math expression evaluator using Python's eval() and math."""
+
+import math
+
+__version__ = "1.0"
+__author__ = "Leodanis Pozo Ramos"
+
+ALLOWED_NAMES = {
+    name: obj for name, obj in math.__dict__.items() if "__" not in name
```
bryanwweber: Same comment as on the article, whether this should be `if not name.startswith("__")`. Make sure to keep them in sync

python-eval-mathrepl/mathrepl.py
```diff
@@ -0,0 +1,78 @@
...
+}
+
+PS1 = "mr>>"
+
+WELCOME = f"""
+MathREPL {__version__}, your Python math expressions evaluator!
+Enter a valid math expression after the prompt "{PS1}".
+Type "help" for more information.
+Type "quit" or "exit" to exit.
```
bryanwweber: If someone types (literally) `"quit"`, then that string will be printed to the output. I'm not sure what the best way to resolve this is... Maybe it is fine to just assume people will know what to do.

python-eval-mathrepl/mathrepl.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+
+USAGE = f"""
+Usage:
+Build math expressions using numeric values and operators.
+Use any of the following functions and constants:
+
+{', '.join(ALLOWED_NAMES)}
```
bryanwweber: Same comment as the article, `ALLOWED_NAMES.keys()`?

python-eval-mathrepl/mathrepl.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+
+
+def evaluate(expression: str) -> float:
```
bryanwweber: Same comment on this line about type hints

python-eval-mathrepl/mathrepl.py
```diff
@@ -0,0 +1,78 @@
...
+            print(USAGE)
+            continue
+        if expression.lower() in {"quit", "exit"}:
+            raise SystemExit()
+
+        # Evaluate the expression
+        try:
+            result = evaluate(expression)
+        except Exception as err:
+            print(err)
```
bryanwweber: Ah, I didn't catch this one on the article.  You should add `continue` below the `print(err)` here, to avoid printing the previous result (which may be undefined on the first loop iteration)

lpozo: My mistake, I'll fix this.

lpozo:
Hey @bryanwweber, I've finish the TR update. Please, let me know if I need to update something else.

python-eval-mathrepl/mathrepl.py
bryanwweber: @lpozo This isn't fixed here

lpozo: I moved the print call to the `try` block, I think this fixes the problem, don't you?

bryanwweber: This is where it was originally. The problem is that if there is an exception, the error is printed and then the script runs the next line, which is to print the result. If the result is undefined, it results in another, unhandled, error. If the result is defined from a previous successful calculation, the error message is printed and then the success message is printed.

lpozo: Yes, but that code isn't there anymore. We don't have any statement below `print(err)` in this new version. Do you think that we still have the same problem?

bryanwweber: Oh, I see you moved the other print call inside the `try` block. Well, it definitely won't have the same problem, but I'm a little torn about that. Ideally, you'd do as little as possible inside the `try` statement, to make sure that you only raised errors you wanted to handle. On the other hand, you're deliberately catching all errors in this block on purpose...

After some thought, I think it would be better to have the `print("The result is...")` outside the `try` block. The reason is that if `print()` encounters an error for some reason, you don't want to be catching that error, the user should definitely know if `print()` is having a problem, and it should probably "crash" the program. In addition, having a separate print call for the result is somehow **clearer** to me, it is more obvious what is being done in each step for the reader. Does that all make sense? If you strongly disagree, feel free to leave the printing of the result in the `try` block.

lpozo:
Hey @bryanwweber, I've rework the example. Please, take a look at it and let me know your opinion. Thanks.

PR #111, opened 2020-05-02
FIX: ma.ModelSchema removed in Flask-Marshmallow 0.12.0

Erff:
Replaced ma.ModelSchema by ma.SQLAlchemyAutoSchema
https://github.com/marshmallow-code/flask-marshmallow/blob/dev/CHANGELOG.rst#0120-2020-04-26


flask-connexion-rest-part-2/version_1/models.py
```diff
@@ -12,7 +12,15 @@ class Person(db.Model):
...
+# class PersonSchema(ma.ModelSchema):
+#    class Meta:
+#        model = Person
+#        sqla_session = db.session
+
+# flask-marshmallow>=0.12.0 (recommended)
+class AuthorSchema(ma.SQLAlchemyAutoSchema):
     class Meta:
         model = Person
         sqla_session = db.session
```
jackchi, line 26: Is `sqla_session` still needed?

Erff: Actually no, it is not needed. I just pushed a fix that leaves the example working.

SouthernYoda:
Thanks for posting a fix!

PR #113, opened 2020-05-06
Heapq

moshez:
**Where to put new files:**

4. 🎉


moshez:
Hi @somacdivad -- I don't have the solutions off hand, but I know how I would approach them (the first couple are just changing `get_neighbors()`, the second one requires keeping "current energy" in the `tentative` dictionary and updating it with either `-1` or `+10` depending on the value, the third would change how path "length" calculation would change, again requiring keeping it in the `tentative` dictionary).

I decided against putting these hints in because I feel they would "spoil" too much of the problem: the point is to struggle with the code to modify it and, in the process, understand it.

somacdivad:
@moshez Just a heads up we realized that this didn't get merged and included in your heapq article as an optin. I just did a quick pass to make sure the code matches what you've got in the article. @jaschmitt will be merging this and including the optin link in the article for you. Nothing needed from you, just wanted to give you a heads up.

PR #121, opened 2020-05-26
'Deploy a Django App With Gunicorn, Nginx, & HTTPS' companion repo

bsolomon1124:
**ATTENTION**: Please squash commits when merging.


django-gunicorn-nginx/README.md
```diff
@@ -0,0 +1 @@
+# Deploy a Django App With Gunicorn, Nginx, & HTTPS
```
bzaczynski, line 1: @bsolomon1124 I believe we want to include a short description of what this is and some instructions for the readers.

bzaczynski (changes requested):

django-gunicorn-nginx/README.md
bsolomon1124: Added in 95e32b1.

bzaczynski (approved):
@bsolomon1124 LGTM :+1:

PR #128, opened 2020-06-15
Material for Python import article

gahjelle:
Article:
- https://github.com/realpython/tutorial-drafts/pull/256
- https://realpython.com/python-import/

**Where to put new files:**

4. 🎉


python-import/cyclical_imports/yang.py
```diff
@@ -0,0 +1,14 @@
+# yang.py
+
+print("Hello from yang")
+import yin  # noqa
```
somacdivad, line 4: Is there a reason you're calling `print()` before the `import`? If not, we could swap the order here so that you don't need to suppress the linter. The `# noqa` comment might be confusing...

python-import/finders_and_loaders/blacklist_importer.py
```diff
@@ -0,0 +1,15 @@
+# blacklist_importer.py
+
+import sys
+
+BLACKLISTED_MODULES = {"re"}
```
somacdivad, line 5: In light of recent events, I've seen some people speaking out against the allusions to race in the whitelist/blacklist terminology.

With things as heated as they are in the US, I wouldn't be surprised if someone commented on it.

How do you feel about replacing the word "Blacklisted" with "Banned"?

python-import/cyclical_imports/yang.py
gahjelle: This is part of a cyclical import example, and the `print()` is there to make it clear when each module is imported. So one of the points of the example is that `print()` is before `import`. See https://realpython.com/python-import/#handle-cyclical-imports

One option could be to add some more explanation to `# noqa`:

```
import yin  # noqa  Import should normally be on top of file
```

Another option could be to remove this example from `materials`. I think many of the other examples are useful to offer as "finished" code, but this one doesn't do much useful at the end and the code really only makes sense if you follow along in the tutorial from the first version to this last one.

python-import/finders_and_loaders/blacklist_importer.py
gahjelle: Makes sense. Thanks for picking up on that.

I've updated the file in `materials`. @jaschmitt Should I do the corresponding updates in the article itself, or will you handle that?

- Rename file: `blacklist_importer.py` -> `ban_importer.py`
- Rename constant: `BLACKLISTED_MODULES` -> `BANNED_MODULES`
- Rename class: `BlacklistFinder` -> `BanFinder`
- Rephrase error message: `f"{name!r} is blacklisted"` -> `f"{name!r} is banned"`

PR #130, opened 2020-06-27
Fix illegal list element raising wrong Exception

Fittiboy:
The "List is empty" Exception was raised if `self.head` was manually changed to 0.
This fix changes `if not self.head` to `if self.head is None` to raise the correct exception.


dbader:
Thanks, that makes sense. LGTM! I also updated the article at https://realpython.com/linked-lists-python/

PR #132, opened 2020-07-07
Create Data Version Control tutorial materials

KristijanIvancic:
**Where to put new files:**

4. 🎉


python-data-version-control/src/evaluate.py
```diff
@@ -0,0 +1,29 @@
...
+from train import load_images, load_labels, preprocess
+
+
+def main(repo_path):
+    test_csv_path = repo_path / "data/prepared/test.csv"
+    test_df = pd.read_csv(test_csv_path)
+    raw_images = load_images(data_frame=test_df, column_name="filename")
+    labels = load_labels(data_frame=test_df, column_name="label")
+    preprocessed = [preprocess(image) for image in raw_images]
+    test_data = np.concatenate(preprocessed, axis=0)
```
somacdivad, line 17: I'm wondering if it would make sense to refactor this out into a function called `load_data()` that returns a tuple of containing the list of labels and the processed images? Something like:

```python
def load_data(data_path):
    df = pd.read_csv(data_path)
    labels = load_labels(data_frame=df, column_name="label")
    raw_images = load_images(data_fram=df, column_name="filename")
    processed_images = [preprocess(image) for image in raw_images]
    data = np.concatenate(processed_images, axis=0)
    return labels, data
```

Then you could just call the function in `main()`.

That might make it easier to describe what's going on when you do your line-by-line breakdown in the article.

Just a thought!

python-data-version-control/src/evaluate.py
```diff
@@ -0,0 +1,29 @@
...
+    model_path = repo_path / "model"
+    model = load(model_path / "model.joblib")
```
somacdivad, line 19: Splitting this across two lines doesn't feel super necessary to me, especially since you don't use `model_path` a second time:

```suggestion
    model = load(repo_path / "model/model.joblib")
```

python-data-version-control/src/evaluate.py
```diff
@@ -0,0 +1,29 @@
...
+    predictions = model.predict(test_data)
+    accuracy = accuracy_score(labels, predictions)
+    metrics = {"accuracy": accuracy}
+    accuracy_path = repo_path / "metrics/accuracy.json"
+    accuracy_path.write_text(json.dumps(metrics))
+
+
+if __name__ == "__main__":
+    repo_path = Path(__file__) / "../.."
+    main(repo_path.resolve())
```
somacdivad, line 29: You can get to the repo directory using the `.parent` attribute twice, and then you don't have to call `.resolve()`:

```suggestion
    repo_path = Path(__file__).parent.parent
    main(repo_path)
```

That seems less hacky to me than using `"../.."` and `.resolve()`. I suggest we change that in all of the relevant files.

python-data-version-control/src/prepare.py
```diff
@@ -0,0 +1,43 @@
+from pathlib import Path
+import pandas as pd
+
+folders_to_labels = {"n03445777": "golf ball", "n03888257": "parachute"}
```
somacdivad, line 4: Let's make this uppercase since we're using it as a constant:

```suggestion
FOLDERS_TO_LABELS = {"n03445777": "golf ball", "n03888257": "parachute"}
```

python-data-version-control/src/prepare.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def get_filelist_and_labels(source_path, target_folders):
```
somacdivad, line 7: It feels a little odd to me that you're taking `target_folders` as an argument here. It seems like you would just check that the folder for an image is in `folders_to_labels` (on line 13) are and not worry about taking any argument.

I guess there might be some architectural reason for doing something like this in a real-world project, but for this example, I think we can simplify it.

```suggestion
def get_filelist_and_labels(source_path):
```

python-data-version-control/src/prepare.py
```diff
@@ -0,0 +1,43 @@
...
+    images = []
+    labels = []
+    for image_path in source_path.rglob("*/*.JPEG"):
+        filename = image_path.absolute()
+        folder = image_path.parts[-2]
```
somacdivad, line 12: Instead of subscribing the `.parts` list, it might be more readable to use `.parent.name`:

```suggestion
        folder = image_path.parent.name
```

python-data-version-control/src/prepare.py
```diff
@@ -0,0 +1,43 @@
...
+        if folder in target_folders:
+            images.append(filename)
+            label = folders_to_labels[folder]
```
somacdivad, line 15: This is just to sync the `if` statement with other updates:

```suggestion
        if folder in FOLDERS_TO_LABELS:
            images.append(filename)
            label = FOLDERS_TO_LABELS[folder]
```

python-data-version-control/src/prepare.py
```diff
@@ -0,0 +1,43 @@
...
+def main(repo_path):
+    data_path = repo_path / "data"
+    train_path = data_path / "raw/train"
+    test_path = data_path / "raw/val"
+    train_filenames, train_labels = get_filelist_and_labels(
+        train_path, folders_to_labels.keys()
+    )
+    test_filenames, test_labels = get_filelist_and_labels(
+        test_path, folders_to_labels.keys()
+    )
```
somacdivad, line 35: If you do remove the `target_folders` parameter from `get_filelist_and_labels()`, then you can shorten these lines:

```suggestion
    train_filenames, train_labels = get_filelist_and_labels(train_path)
    test_filenames, test_labels = get_filelist_and_labels(test_path)
```

python-data-version-control/src/train.py
```diff
@@ -0,0 +1,42 @@
...
+    return reshaped
+
+
+def main(repo_path):
+    train_csv_path = repo_path / "data/prepared/train.csv"
+    train_df = pd.read_csv(train_csv_path)
+    raw_images = load_images(data_frame=train_df, column_name="filename")
+    labels = load_labels(data_frame=train_df, column_name="label")
+    preprocessed = [preprocess(image) for image in raw_images]
+    train_data = np.concatenate(preprocessed, axis=0)
```
somacdivad, line 33: Ah, I just realized that if you refactor this stuff out to the `load_data()` function, then you can just import it in the `evaluate.py` script which will make it even shorter!

python-data-version-control/src/evaluate.py
```diff
@@ -0,0 +1,29 @@
+from joblib import load
+import json
+from pathlib import Path
+import numpy as np
+import pandas as pd
+from sklearn.metrics import accuracy_score
```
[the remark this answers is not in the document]
KristijanIvancic: Heh, I do this usually (and automatically, with `isort`) but decided to remove the empty spaces because I thought it would be more in line with the style guide. Sometimes it's tough to find the balance between standard coding advice and the RP style guide. 😅

python-data-version-control/src/evaluate.py
KristijanIvancic: Sounds good, done!

python-data-version-control/src/evaluate.py
KristijanIvancic: Yep, that does seem less hacky!

python-data-version-control/src/evaluate.py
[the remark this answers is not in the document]
somacdivad: Yeah, I know it can be a little confusing to know where to deviate from PEP 8 and where to follow it. But that's one of the things TR is for! To help you sort all that out!

somacdivad:
@KristijanIvancic I've approved the PR. No need to merge it or anything. An editor will take care of that before the article goes live!

PR #141, opened 2020-07-28
Materials for Consuming APIs article

pedropregueiro:
**Where to put new files:**

4. 🎉


JFincher42 (changes requested):
Hi Pedro:

I checked all the code, and it works great!

While I don't have anything specific to call out in any of these programs, I have some overall observations I'd like you to consider before checking in:

- All of the code needs comments, especially when you are importing libraries other than `requests`. I would also add them for each meaningful block of code.
- You should use f-strings for your output formatting.
- For data like API keys or client keys which the reader should be replacing, your comment should be more noticeable. I would capitalize part of it to draw attention to it.

The GitHub code is a great example of all these. Comments can help explain why there are three different helper functions. All the output is more understandable when formatted with f-strings as well. And since the user needs to get an API key and take some action, those areas should be more noticeable.

Look forward to the next iteration!

--Jon

pedropregueiro:
This one is ready for another pass @JFincher42.

pedropregueiro:
Fixed all remaining linting issues.

JFincher42 (approved):
Love the changes - ship it!

PR #144, opened 2020-08-10
Add tictactoe project files

bmwant:
**Where to put new files:**

4. 🎉

PR for the article itself is [here](https://github.com/realpython/tutorial-drafts/pull/601)


bmwant:
@somacdivad could you review this?

tictactoe/io.py
```diff
@@ -0,0 +1,75 @@
...
+    def print_winner(self, name=None):
+        pass
+
+    @abstractmethod
+    def get_input(self):
+        pass
+
+
+class ConsoleFrontend(IOFrontend):
+    placeholders = ["①", "②", "③", "④", "⑤", "⑥", "⑦", "⑧", "⑨"]
```
somacdivad, line 23: I'm concerned that we don't have a basic frontend that will work on everyone's terminal. Not all terminals will display emojis. I think it would be really nice to have a simple frontend that just worked for everyone, then a fancier frontend (like the existing `ConsoleFrontend`) to illustrate how to plug in different frontends.

Here's what I'm thinking as a super simple frontend:

```python
class BasicConsoleFrontend(IOFrontend):
    def print_board(self, board):
        for i, row in enumerate(board):
            for j, column in enumerate(row):
                if cell != Cell.EMPTY:
                    print(cell, end = "|")
                else:
                    cell_number = i * len(row) + j + 1
                    print(cell_number)
            print()
        print()

    def print_winner(self, name=None):
        if name is None:
            print("Game is a draw!")
        else:
            print(f"Player {name} wins!")

    def get_input(self):
        return input("Enter a number of the cell: ")
```

You could also simplify the nested loop a bit:

```python
for i, row in enumerate(board):
    row_display = []
    for j, col in enumerate(row):
        if cell != Cell.EMPTY:
            row_display.append[cell]
        else:
            cell_number = i * len(row) + j + 1
            row_display.append[cell_number]
    print(" | ".join(row_display)
```

That changes the output a bit so that there isn't the | on the far right-hand side of the table, and avoids the two calls to `print()` without any arguments.

So lots of food for thought there. Curious to know what you think!

tictactoe/io.py
```diff
@@ -0,0 +1,75 @@
...
+        click.clear()
+        for i, row in enumerate(board):
+            table_row = []
+            for j, column in enumerate(row):
+                if (cell := board[i][j]) == Cell.X:
+                    text = click.style(cell.value, fg="red", bold=True)
+                elif cell == Cell.O:
+                    text = click.style(cell.value, fg="blue", bold=True)
+                else:
+                    text = str(i * len(row) + j + 1)
```
somacdivad, line 59: Here's another spot where we could simplify things a little bit:

```python
cell = board[i][j]
if cell != Cell.EMPTY:
    color = "red" if cell == Cell.X else "blue"
    text = click.style(cell.value, fg=color, bold=True)
else:
    text = str(i * len(row) + j + 1)
```

It might even be a good idea to refactor some of this into it's own function, but I'll let you decide on that...

tictactoe/__init__.py
```diff
@@ -0,0 +1,7 @@
+from enum import Enum
```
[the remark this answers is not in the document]
bmwant: yes, but `player.py` also needs `Cell.EMPTY` check, so to prevent circular import we need it to be stored in a "shared" place or having an import within a `get_turn` method which I think would be a bad practice. I might also move this to separate `cell.py` module in case `__init__.py` is not the right place.

tictactoe/game.py
```diff
@@ -0,0 +1,90 @@
+from typing import Optional
+
+from tictactoe import Cell
+from tictactoe.player import RandomPlayer
+from tictactoe.io import IOFrontend
+
+
+class Game(object):
+    def __init__(
+        self, x_player=None, o_player=None, frontend: IOFrontend = None
```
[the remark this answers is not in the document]
bmwant: Agree that type hint here seems like redundant

tictactoe/game.py
```diff
@@ -0,0 +1,90 @@
...
+        for row in self.board:
+            for cell in row:
+                if cell == Cell.EMPTY:
+                    return False
+        return True
+
+    def _check_winner(self) -> Optional[Cell]:
+        # Check rows
+        for row in self.board:
+            if len(set(row)) == 1 and row[0] != Cell.EMPTY:
```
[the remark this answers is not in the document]
bmwant: I've changed code to have separate `_check_winning_set` with shared logic, but I think it's a bit better to have explicit blocks highlighting which `group` we are checking at the moment even with some extra code duplication.

tictactoe/game.py
```diff
@@ -0,0 +1,90 @@
...
+        winner = self._check_winner()
+        if winner is not None:
+            return winner
+
+        return self._check_draw()
+
+    def make_turn(self, turn: int, piece: Cell):
+        size = len(self.board)
+        i = turn // size
+        j = turn % size
```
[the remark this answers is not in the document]
bmwant: created `Turn` class to deal with that

tictactoe/io.py
bmwant: I believe every modern terminal/shell should support proper unicode. Anyway agree to simplify this to just ASCII. The point of doing that initially was a comment from @gahjelle to have some nicer UI.

bmwant: Screen Shot 2020-08-25 at 10 57 12 AM

tictactoe/__init__.py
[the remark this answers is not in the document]
somacdivad: Yeah, I missed that it was also used in `player.py` (left the comment before I had read that file)

tictactoe/io.py
somacdivad: @gahjelle Does love his pretty terminal UIs 🙂

> I believe every modern terminal/shell should support proper Unicode.

Not all readers use "modern" terminals, though. I think having ASCII as a default is the best option. It still looks good and is pretty much guaranteed to work anywhere.

tictactoe/player.py
```diff
@@ -2,18 +2,25 @@
...
+
+
 class Player(ABC):
     def __init__(self, name=None, frontend=None):
         self.name = name
         self.frontend = frontend
 
     @abstractmethod
-    def get_turn(self, board) -> int:
+    def get_turn(self, board) -> Turn:
```
somacdivad, line 23: Since we agreed to remove type hints, I thinkwe can remove this, too:

```suggestion
    def get_turn(self, board):
```

tictactoe/player.py
```diff
@@ -25,21 +32,24 @@ def __init__(self):
...
+        return choice(available_turns)
 
 
 class ConsolePlayer(Player):
     def __init__(self, name="Console Player"):
         frontend = ConsoleFrontend()
         super().__init__(name=name, frontend=frontend)
 
-    def get_turn(self, board) -> int:
+    def get_turn(self, board) -> Turn:
```
somacdivad, line 49: Here's another type hint that snuck through 🙂

```suggestion
    def get_turn(self, board):
```

bmwant:
hey @somacdivad, I've moved all the code to subfolder to be able to invoke it via `python -m`. Initially you supposed to run it from parent directory and this way is described in the article but here I think it's ok to have inner folder. Also I have added a `requirements.txt` file and instructions to the `README.md` file (creation of virtual environment is also described in the article itself).

PR #170, opened 2020-11-15
Python web applications

martin-martin:
**Where to put new files:**

4. 🎉


python-web-applications/requirements.txt
```diff
@@ -0,0 +1 @@
+flask
```
bzaczynski, line 1: 👩‍💻 **Technical Recommendation**

We should be instilling good programming practices in our readers, so let's specify a Flask version. That will mitigate potential problems with breaking changes in Flask when someone reads this article in the future. We've had a similar situation in a Django article before.

```suggestion
Flask==1.1.2
```

python-web-applications/main.py
```diff
@@ -0,0 +1,35 @@
...
+                Celsius: <input type="text" name="cel_temp">
+                <input type="submit" value="Convert">
+            </form>"""
+        + "Fahrenheit: "
+        + far_temp
+    )
+
+
+if __name__ == "__main__":
+    app.run(host="127.0.0.1", port=8080, debug=True)
```
bzaczynski, line 35: It's great that you use the standard black formatting for the supplementing code, unlike the code that we show in the article. However, I'd use more descriptive names and drop a few comments:

```suggestion
from flask import Flask
from flask import request

app = Flask(__name__)


def fahrenheit_from(celsius):
    """Convert Celsius to Fahrenheit degrees."""
    try:
        fahrenheit = float(celsius) * 9 / 5 + 32
        fahrenheit = round(fahrenheit, 3)  # Round to three decimal places
        return str(fahrenheit)
    except ValueError:
        return "invalid input"


@app.route("/")
def index():
    celsius = request.args.get("celsius", "")
    if celsius:
        fahrenheit = fahrenheit_from(celsius)
    else:
        fahrenheit = ""
    return (
        """<form action="" method="get">
                Celsius temperature: <input type="text" name="celsius">
                <input type="submit" value="Convert to Fahrenheit">
            </form>"""
        + "Fahrenheit: "
        + fahrenheit
    )


if __name__ == "__main__":
    app.run(host="127.0.0.1", port=8080, debug=True)

```

bzaczynski (changes requested):
@martin-martin Please take a look at my comments. Thanks!

python-web-applications/main.py
martin-martin: sure, I agree with that. I left most of the code the way it was, but why not also update the code example, not just the text : )

I'll propagate the changes into the code snippets in the article as well.

bzaczynski (approved):
lgtm :+1:

PR #171, opened 2020-11-17
Update sentiment_analyzer.py

kylestratis:
Published article material fix @somacdivad


somacdivad (approved):
LGTM 👍

Out of curiosity @kylestratis, was this something that just didn’t get updated on materials after TR, or did you have to fix this on the article as well?

PR #182, opened 2021-01-28
[BUGFIX] Reference the global PEOPLE dict in create in people.py of connexion-v3

rootAvish:
The creation is not persisted otherwise outside the function.

How to reproduce the associated bug:

- Use the API to create new record through its POST method.
- Use the people/{lname} API to fetch the newly added record.
- Nothing will be returned since the global `PEOPLE` dict has not been updated.

### DIFF
--- a/flask-connexion-rest/version_3/people.py
+++ b/flask-connexion-rest/version_3/people.py
@@ -74,6 +74,7 @@ def create(person):
     :param person:  person to create in people structure
     :return:        201 on success, 406 on person exists
     """
+    global PEOPLE
     lname = person.get("lname", None)
     fname = person.get("fname", None)
 


flask-connexion-rest/version_3/people.py
```diff
@@ -74,6 +74,7 @@ def create(person):
     :param person:  person to create in people structure
     :return:        201 on success, 406 on person exists
     """
+    global PEOPLE
```
dbader, line 77: That shouldn't be necessary since we're not assigning to `PEOPLE` in this function (`PEOPLE = ...`). Instead, the existing `PEOPLE` dict gets updated: `PEOPLE[lname] = { ... `

rootAvish: @dbader, here's the same behaviour reproduced on a micro level, it does not get updated:

```
In [9]: people = {'gupta': 'avishkar'}

In [10]: def update():
    ...:     people['sharma'] = 'avishkar'
    ...:

In [11]: people
Out[11]: {'gupta': 'avishkar'}
```

PR #183, opened 2021-02-12
Directory Tree Generator source code

lpozo:
**Where to put new files:**

4. 🎉


directory-tree-generator-python/README.md
```diff
@@ -0,0 +1,44 @@
...
+(venv) $
+```
+
+2. Run the application
+
+```sh
+(venv) $ python tree.py /path/to/directory/
+```
+
+**Note:** The `-h` or `--help` option provides help on how to use RP Tree.
```
somacdivad, line 24: The only thing I would add here is that the `hello/` directory is included here as a sample directory, and I think it would be a good idea to include the example of running the tool on `hello/` and showing the terminal output.

somacdivad (changes requested):
Hey @lpozo! Looks really good! Just one tiny update I think we should make here. See my comment below.

somacdivad (approved):
Looks good!

PR #187, opened 2021-03-05
Complete code for embedded Python article

notexactlyawe:
**Where to put new files:**

4. 🎉


notexactlyawe:
Linter is failing because of the `from microbit import *`, however that's idiomatic when writing code for the micro:bit. See the [accelerometer examples in the docs](https://microbit-micropython.readthedocs.io/en/v1.0.1/accelerometer.html#examples) for an example.

dbader:
Thanks @notexactlyawe -- I agree it makes sense to keep the wildcard around in this case. To pass the linter check, you can add a `# noqa` comment on the affected line:

```python
from microbit import *  # noqa
from random import randrange
...
```

notexactlyawe:
Thanks @dbader. It looks like that fixes the error on line 1 but not the other errors that come when I use modules/functions from `microbit`. I can:
 - Add `noqa` comments to all affected lines
 - Add a file-level noqa (which would squash any other errors that appear if we change this file)
 - Ignore the CI errors

Do you have a preferred fix?

dbader:
Hmm...so, thinking about this again, maybe it would actually be better to not do the wildcard import then. Sure it's what the microbit docs do, but there should be no downside/difference to using named imports. And it would clean up the code...

notexactlyawe:
Thanks @dbader, I've expanded the import and the linter is passing now. I'll come back and update the README when we have a finalized title and link to the article.

bzaczynski:
lgtm

jaschmitt (approved):
Changes implemented by author

PR #198, opened 2021-04-19
Sample code for the Complex Numbers tutorial

bzaczynski:
**Where to put new files:**

4. 🎉


somacdivad (changes requested):
These are really cool notebooks @bzaczynski!

I found one small issue in the `bermuda.ipynb` file where you used the word **negate** to mean "take the negative of", which we removed from the article. We should update that here, too.

Also, I would normally argue against using `global` like you have in the `fourier.ypynb` file, but I understand why you did it in this context. I wonder if we should add a note in a Markdown cell explaining that although this is typically considered poor practice, there is a practical reason you're doing it here. What do you think?

bzaczynski:
@somacdivad Thanks for catching this. I added a Jupyter Notebook's alert box to warn about the global keyword and changed the README file as there's no need to install Pillow anymore.

somacdivad (approved):
Thanks for the updates @bzaczynski! Everything looks good!

PR #206, opened 2021-05-28
Update schema to use class reference instead of string

daneah:
Correcting a line after feedback from a reader


dbader (approved):
Thanks @daneah! LGTM! :)

PR #213, opened 2021-07-30
add C++ examples for ctypes

adam-dziedzic:
**Where to put new files:**

- New files should go into a top-level subfolder, named after the article slug. For example: `my-awesome-article`

**How to merge your changes:** 

1. [Make sure the CI code style tests all pass (+ run the automatic code formatter if necessary).](https://github.com/realpython/materials/blob/master/README.md)
2. Find an RP Team member on Slack and ask them to review & approve your PR.
3. Once the PR has one positive ("approved") review, GitHub lets you merge the PR.
4. 🎉

PR #215, opened 2021-08-19
Fixing error in tasks.py causing failures

jima80525:
This is fixing an error I missed on a PR yesterday from an interested reader.  Lesson learned - I should have tested it out before approving and not just relying on visual inspection.  Apologies!

### DIFF
--- a/python-bindings/tasks.py
+++ b/python-bindings/tasks.py
@@ -64,7 +64,7 @@ def build_cmult(c, path=None):
 
 
 @invoke.task()
-def test_ctypes():
+def test_ctypes(c):
     """Run the script to test ctypes"""
     print_banner("Testing ctypes Module for C")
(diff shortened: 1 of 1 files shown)


gahjelle:
The issue is the missing context parameter, right? As mentioned here: http://docs.pyinvoke.org/en/stable/getting-started.html#defining-and-running-task-functions

My guess is that it was removed in #213 because of a confusion as to what the name `c` was referring to? Which I guess is a bit particular for this repo since it deals with C code ...

I think this looks good. My one suggestion would be to use a more verbose name, `ctx` or even `context`, to avoid similar misunderstandings down the road.

dbader:
Alright, thanks for the hotfix there :) I'll merge this now to fix the sample code (+1 to what @gahjelle suggested re: naming)

dbader (approved):

PR #221, opened 2021-09-19
Mandelbrot

bzaczynski:
**Where to put new files:**

4. 🎉


mandelbrot-set-python/mandelbrot_03.py
```diff
@@ -0,0 +1,27 @@
...
+    escape_radius: float = 2.0
+
+    def __contains__(self, c: complex) -> bool:
+        return self.stability(c) == 1
+
+    def stability(self, c: complex, smooth=False, clamp=True) -> float:
+        value = self.escape_count(c, smooth) / self.max_iterations
+        return max(0.0, min(value, 1.0)) if clamp else value
+
+    def escape_count(self, c: complex, smooth=False) -> int | float:
```
gahjelle, line 19: Out of the box, the `int | float` syntax means that this only runs on Python 3.10+. However, we can add a `from __future__ import annotations` on the top of this file to make the syntax supported on older versions, and this module work on Python 3.7+

mandelbrot-set-python/README.md
```diff
@@ -0,0 +1,26 @@
...
+Code snippets and Python scripts supplementing the [Draw the Mandelbrot Set in Python](https://realpython.com/mandelbrot-set-python) tutorial on [Real Python](https://realpython.com/).
+
+## Installation
+
+This code requires the following libraries:
+
+- Matplotlib
+- NumPy
+- Pillow
+- SciPy
```
gahjelle, line 12: Since you're listing these explicitly, I think it would be more helpful to use the exact PyPI names (so basically change to lowercase) as this would allow any power reader to install manually instead of using the prepared and locked `requirements.txt`:

```suggestion
- `matplotlib`
- `numpy`
- `pillow`
- `scipy`
```

gahjelle (commented):
@bzaczynski I think looks great! My main comment is that it would be good to hide union type annotation being a future import to make the code runnable on Python versions before 3.10.

mandelbrot-set-python/README.md
bzaczynski: These are the proper names of the libraries I'm referring to. PyPI is case insensitive, so it will accept uppercase, lowercase, and everything in between.

mandelbrot-set-python/mandelbrot_03.py
bzaczynski: Nice trick! I haven't been using `from __future__ import ...` and almost forgot about its existence. Thanks :wink:

gahjelle (approved):
@bzaczynski I did a final test of the code, and everything still runs beautifully :+1: 

I'll be merging this in so we're ready to set up the tutorial.

PR #222, opened 2021-09-20
Fix inconsistent variable name in for-loop

acsany:
In `django-diary/django-diary/source_code_*/entries/templates/entries/entry_list.html` there was an inconsistency in using `object_list` instead of `entry_list`. The fixes are already live in the CMS. This pull request updates the source code in the Materials repo accordingly.


martin-martin (approved):
Perfect, thanks for fixing this @acsany

PR #228, opened 2021-10-26
Python Dice Roll SbSP

lpozo:
**Where to put new files:**

4. 🎉


python-dice-roll/source_code_final/dice.py
```diff
@@ -0,0 +1,189 @@
...
+
+# def _generate_dice_faces_rows(dice_faces):
+#     dice_faces_rows = []
+#     for row_idx in range(DIE_HEIGHT):
+#         row_components = []
+#         for die in dice_faces:
+#             row_components.append(die[row_idx])
+#         row_string = DIE_FACE_SEPARATOR.join(row_components)
+#         dice_faces_rows.append(row_string)
+#     return dice_faces_rows
```
martin-martin, line 177: I think it's awkward to have this commented-out code in the source code, especially in so many different steps.

IMO this could get resolved if you move the refactoring into its own step (Step 4) instead of having it as an optional exercise in step 3.

I've mentioned this in more detail over in the tutorial-drafts PR: https://github.com/realpython/tutorial-drafts/pull/744/files#diff-1a870bf5e170073f6e7784c862223407c720d270c35f8d181d2e14cbe655cf9aR398, but I can see immediate advantages of going down that road in how it'd relate to the code in the source-code folders here.

python-dice-roll/source_code_step_3/dice.py
```diff
@@ -0,0 +1,170 @@
...
+
+    dice_faces_diagram = "\n".join([diagram_header] + dice_faces_rows)
+    return dice_faces_diagram
+
+
+# EXERCISE:
+# Refactor generate_dice_faces_diagram() to satisfy the single-responsibility
+# principle. Take advantage of _get_dice_faces() and write another helper
+# function called _generate_dice_faces_rows() to extract the functionality
+# from lines 115 to 121.
```
martin-martin, line 120: In case you don't agree with the step 4 refactoring mentioned elsewhere, then you'll have to make sure that the line numbers here make sense. Currently these commented-out instructions about refactoring point to lines 115 to 121, which isn't correct.

martin-martin (commented):
Hi @lpozo , thanks for adding the source code folders.

I have a suggestion about refactoring the final step of your tutorial to move it from a quite short section on just adding two client code calls, to be about the refactoring instead, which is explained more in the [tutorial drafts PR](https://github.com/realpython/tutorial-drafts/pull/744#pullrequestreview-804921945)

martin-martin (approved):
This looks good to me. There's one relevant suggestion in the tutorials-drafts PR that you might want to consider, but no biggie: https://github.com/realpython/tutorial-drafts/pull/744#discussion_r758192865 :)

PR #247, opened 2022-02-09
Sockets update to f strings and copy instead of list

iansedano:
**Where to put new files:**

4. 🎉

### DIFF
--- a/python-sockets-tutorial/echo-server.py
+++ b/python-sockets-tutorial/echo-server.py
@@ -10,7 +10,7 @@ with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
     s.listen()
     conn, addr = s.accept()
     with conn:
-        print("Connected by", addr)
+        print(f"Connected by {addr}")
         while True:
             data = conn.recv(1024)
             if not data:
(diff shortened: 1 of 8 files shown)


python-sockets-tutorial/echo-server.py
```diff
@@ -10,7 +10,7 @@
     s.listen()
     conn, addr = s.accept()
     with conn:
-        print("Connected by", addr)
+        print(f"Connected by {addr}")
```
gahjelle, line 13: Most other printouts start with a lower-case letter, so we could change this to be consistent.

However, I'm leaning towards suggesting switching the others to start with an upper-case letter, as that would better reflect our other practices at Real Python.

gahjelle (approved):
Great job, @iansedano 

I added a few suggestions. Most of them are about using the `!r` formatting directive instead of calling `repr()` (https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals, https://realpython.com/python-f-strings/#arbitrary-expressions). It would be great to showcase this little-known feature :)

I'm also wondering if it might be worth it to upper-case the first character in the text... but we should also know when to stop :yum:

gahjelle:
Looks great! :+1:

PR #248, opened 2022-02-14
Add Python Wiki Django Project

acsany:
Material files for the Django Pagination tutorial.


django-pagination/README.md
```diff
@@ -0,0 +1,77 @@
...
+This repository contains the source code for the Django project that is referred to in the [Real Python Django Pagination Article](https://realpython.com/django-pagination/).
+
+## Create the Django Project
+
+After you cloned this repository and navigated into it's folder, you can run the provided example project on your local machine by following the steps outlined below.
+
+Create a new virtual environment:
+
+```bash
+python3 -m venv venv
```
bzaczynski, line 12: :tophat: **Style Tip**

Please remember to include the leading dollar sign:

```suggestion
$ python3 -m venv venv
```

django-pagination/README.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+Activate the virtual environment:
+
+```bash
+source ./venv/bin/activate
```
bzaczynski, line 18: :tophat: **Style Tip**

Please remember to include the leading dollar sign:

```suggestion
$ source ./venv/bin/activate
```

django-pagination/README.md
```diff
@@ -0,0 +1,77 @@
...
+Enter the Django shell:
+
+```bash
+(venv) $ python manage.py shell
+````
+
+Add the Python keywords to your database:
+
+```pycon
+>>> import keywords
```
bzaczynski, line 55: :warning: **Fact Check**

The name of the built-in module is singular:

```suggestion
>>> import keyword
```

django-pagination/Readme.md
```diff
@@ -0,0 +1,77 @@
+# Django Project: Python Wiki
```
bzaczynski, line 1: :warning: **Fact Check**

It looks like the additional `Readme.md` file is redundant as having the default `README.md` should be sufficient.

django-pagination/terms/templates/terms/base.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <title>Python Keywords</title>
+    <link rel="stylesheet" href="https://cdn.simplecss.org/simple.min.css">
```
bzaczynski, line 6: :wrench: **Technical Recommendation**

I'd recommend using a local CSS file instead of relying on an external source, which might go out of service. Besides, readers without Internet access won't be able to run your code and see the expected result.

django-pagination/terms/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.urls import path
+from . import views
+
+urlpatterns = [
+    path(
+        "all", views.AllKeywordsView.as_view(template_name="terms/base.html")
```
bzaczynski, line 6: :warning: **Fact Check**

What about the other views you mentioned in the tutorial? It would be helpful and convenient for the readers to include them here as well.

bzaczynski (changes requested):
Hey @acsany :wave: 

I left a few suggestions for you on the materials PR.

django-pagination/terms/urls.py
acsany: Hmm … this makes me think if I approached the materials repo the wrong way. 🤔 

My idea was to have the project as a starting point. However, maybe it’d be even better to provide the final project with all the examples I show in the tutorial included. What do you think?

django-pagination/terms/templates/terms/base.html
```diff
@@ -0,0 +1,42 @@
...
+
+    {% block content%}
+        {% for keyword in keyword_list %}
+            <pre>{{keyword}}</pre>
+        {% endfor %}
+    {% endblock %}
+
+    <nav>
+        {% block pagination %}
+        <!-- Placeholder: Pagination-Elements -->
```
bzaczynski, line 36: Would it make more sense to use Django Template comments instead of HTML comments here?

```html
{# Placeholder: Pagination-Elements #}
```

django-pagination/terms/templates/terms/infinite_scrolling.html
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        onIntersect() {
+            this.pageIndex++
+            this.loadMore()
+        }
+
+        watchIntersection() {
+            document.addEventListener("DOMContentLoaded", () => {
+                let observer = new IntersectionObserver(this.onIntersect.bind(this),
```
bzaczynski, line 37: It's a common "best practice" to use `const` as much as possible and only use `let` when necessary.

django-pagination/README.md
```diff
@@ -0,0 +1,87 @@
+# Django Project: Python Wiki
```
bzaczynski, line 1: @acsany We'll probably need to update this once the final title is settled.

django-pagination/terms/templates/terms/infinite_scrolling.html
```diff
@@ -0,0 +1,79 @@
...
+
+        onIntersect() {
+            if (!this.lastPage) {
+                this.pageIndex++
+                this.loadMore()
+            }
+        }
+
+        addElement(keyword) {
+            let pre = document.createElement("pre")
```
bzaczynski, line 52: @acsany Prefer `const` over `let`.

bzaczynski (approved):

PR #259, opened 2022-05-04
Add FastAPI URL Shortener

acsany:
**Where to put new files:**

4. 🎉

### DIFF
--- a/fastapi-url-shortener/source_code_final/README.md
+++ b/fastapi-url-shortener/source_code_final/README.md
@@ -1,0 +1,67 @@
+# FastAPI URL Shortener
+
+## Install the Project
+
+1. Create a Python virtual environment
+
+```sh
+$ python -m venv venv
+$ source venv/bin/activate
+(venv) $
+```
+
+2. Install the requirements
+
+```
+(venv) $ python -m pip install -r requirements.txt
+```
+
+## Run the Project
+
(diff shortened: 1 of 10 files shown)


fastapi-url-shortener/source_code_final/README.md
```diff
@@ -0,0 +1,67 @@
+# FastAPI URL Shortener
```
bzaczynski, line 1: Do we need those extra README files in the individual subfolders beside the one in the root folder?

fastapi-url-shortener/source_code_final/requirements.txt
```diff
@@ -0,0 +1,16 @@
+anyio==3.6.0
+asgiref==3.5.1
+click==8.1.3
+decorator==5.1.1
+fastapi==0.75.0
+h11==0.13.0
```
bzaczynski, line 6: I seem to have one extra dependency in my virtual environment, but that might be some Linux-specific thing maybe? :thinking: 

```
greenlet==1.1.2
```

bzaczynski (approved):
@acsany Looks good to me :+1: 

By the way, do we want to include an explicit Step 4 to follow the tutorial? :thinking: It seems redundant as it's also the final part, but I don't know ¯\_(ツ)_/¯

PR #268, opened 2022-06-17
fix: mkdocstrings Python handler updates

martin-martin:
The legacy Python handler doesn't come by default with a mkdocstrings installation anymore.
Instead you need to install the new Python handler explicitly.

The change is reflected in a different line during installation, as well as required some
extra explanation that I added there. There's also a cascading effect that package-level
docstrings recorded in `__init__.py` aren't parsed through `::: package.__init__` but
instead through `::: package` so I've applied that change as well.

**Where to put new files:**

4. 🎉


martin-martin:
Implemented in the CMS @gahjelle @KateFinegan . Closing because GH is not the truth!

PR #282, opened 2022-08-05
Python abs() - Supplemental Materials

bzaczynski:
**Where to put new files:**

4. 🎉

### DIFF
--- a/python-absolute-value/README.md
+++ b/python-absolute-value/README.md
@@ -1,0 +1,3 @@
+# How to Find an Absolute Value in Python
+
+Supplementing materials for the [tutorial on absolute values](https://realpython.com/python-absolute-value/) hosted on Real Python.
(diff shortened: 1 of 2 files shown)


gahjelle (approved):
@bzaczynski I think this looks good.  Simple and to the point. 

My only suggestion is that we might make the examples a tad more descriptive by either:

- use more descriptive function names, say something like `absolute_value_piecewise()`, `absolute_value_geometric()`, `Vector_free()` etc instead of their numbered counterparts,

Or:

- Add a short and to the point docstring to the classes and functions. For example,  `"""Geometrically defined absolute value"""`

This isn't a blocker but I think it could make the examples slightly quicker to grok and resist to connect to the tutorial for the readers.

bzaczynski:
@gahjelle Thanks for the prompt feedback and the approval. I just updated the code.

gahjelle:
Looks great!

PR #284, opened 2022-08-10
Add supplemental material for HTML and CSS article

acsany:
**Where to put new files:**

4. 🎉


html-and-css/README.md
```diff
@@ -0,0 +1,3 @@
+# HTML and CSS for Python Developers
+
+Supplementing materials for the [tutorial on GitHub Copilot](https://realpython.com/html-and-css-for-python-developers/) hosted on Real Python.
```
bzaczynski, line 3: :warning: **Fact Check**

I've been a victim of the same mistake myself: copy and paste :wink: 

```suggestion
Supplementing materials for the [tutorial on HTML and CSS for Python Developers](https://realpython.com/html-and-css-for-python-developers/) hosted on Real Python.
```

bzaczynski (approved):
@acsany Apart from one simple copy-paste mistake, it looks good to me :blush: 

By the way, you haven't included any Python scripts shown in the tutorial. I think it would be helpful to include them here as well!

acsany:
Thanks for your review and for spotting the copy-paste mistake. 😬

> By the way, you haven't included any Python scripts shown in the tutorial. I think it would be helpful to include them here as well

I was thinking about adding the Python files but I refrained from that because of two reasons:

1. The scripts make sense for the tutorial but don't show best practices for creating and parsing HTML. 
2. One idea of the supplementing files is, that they can be used as a project quickstart. Some readers might be confused about the Python files in it.

But I don't feel strongly about it. If you think it would bring value to the readers, I'm happy to add them, too. 🙂

bzaczynski:
@acsany Yeah, I think some readers will appreciate having all the Python code in one place :+1:

PR #291, opened 2022-08-16
Add black formatting to Jupyter Notebooks

gahjelle:
Adds the new feature where `black` can reformat Jupyter Notebooks (https://black.readthedocs.io/en/stable/getting_started.html#installation). This was introduced in v21.8b0.


gahjelle:
Black now checks 1543 files, previously it checked 1519 files. So it's checking the notebooks as well :blush:

dbader (approved):

PR #308, opened 2022-09-26
Add materials for Nim Minimax tutorial

gahjelle:
**Where to put new files:**

4. 🎉


python-minimax-nim/alphabeta_nim.py
```diff
@@ -0,0 +1,38 @@
...
+
+def evaluate(state, is_maximizing):
+    if all(counters == 0 for counters in state):
+        return 1 if is_maximizing else -1
+
+
+def possible_new_states(state):
+    for pile, counters in enumerate(state):
+        for remain in range(counters):
+            yield state[:pile] + (remain,) + state[pile + 1 :]
```
bzaczynski, line 38: You may want to swap the order of these two functions to keep the code consistent with the tutorial.

python-minimax-nim/minimax_nim_split.py
```diff
@@ -0,0 +1,26 @@
+def minimax(state, is_maximizing):
```
bzaczynski, line 1: The tutorial version uses caching:

```python
from functools import cache

@cache
def minimax(state, is_maximizing):
    ...
```

bzaczynski (approved):
@gahjelle Approved with two tiny remarks.

python-minimax-nim/minimax_nim_split.py
gahjelle: Hmm ... I don't think I'm explicit about which file the caching is added to (except for the alpha-beta one). Since this is just the solution of one of the exercises, I haven't even listed the `minimax()` function except to say that it's copied from `minimax_nim.py` and this happens before the caching is introduced in the following section.

That said, it probably makes sense to add the caching on all the files.

gahjelle:
Thanks @bzaczynski

PR #316, opened 2022-10-11
Sample code for the `doctest` article

lpozo:
**Where to put new files:**

4. 🎉

### DIFF
--- a/python-doctest/fizz_buzz.py
+++ b/python-doctest/fizz_buzz.py
@@ -1,0 +1,72 @@
+# Replace numbers that are divisible by 3 with "fizz"
+def fizz_buzz(numbers):
+    """Implement the Fizz buzz game.
+    
+    >>> fizz_buzz([3, 6, 9, 12])
+    ['fizz', 'fizz', 'fizz', 'fizz']
+    """
+
+# # Replace numbers that are divisible by 3 with "fizz"
+# def fizz_buzz(numbers):
+#     """Implement the Fizz buzz game.
+    
+#     >>> fizz_buzz([3, 6, 9, 12])
+#     ['fizz', 'fizz', 'fizz', 'fizz']
+#     """
+#     result = []
+#     for number in numbers:
+#         if number % 3 == 0:
+#             result.append("fizz")
+#         else:
+#             result.append(number)
+#     return result
+
+
+# Replace numbers that are divisible by 5 with "buzz"
+# def fizz_buzz(numbers):
+#     """Implement the Fizz buzz game.
+    
+#     >>> fizz_buzz([3, 6, 9, 12])
+#     ['fizz', 'fizz', 'fizz', 'fizz']
+
+#     >>> fizz_buzz([5, 10, 20, 25])
+#     ['buzz', 'buzz', 'buzz', 'buzz']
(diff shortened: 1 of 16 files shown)


python-doctest/user.py
```diff
@@ -0,0 +1,46 @@
+# class User:
```
iansedano, line 1: Does it make sense to add a comment to say what to do with this file? Or maybe we rename the `User` to `User_one`, `User_two` etc, so that when they run it they can see everything?

python-doctest/fizz_buzz.py
```diff
@@ -0,0 +1,73 @@
...
+#         if number % 3 == 0:
+#             result.append("fizz")
+#         elif number % 5 == 0:
+#             result.append("buzz")
+#         else:
+#             result.append(number)
+#     return result
+
+
+# Replace numbers that are divisible by 3 and 5 with "fizz buzz"
```
iansedano, line 47: I think in this case it makes sense to leave the final implementation uncommented, and leave a comment saying that the commented versions are previous stages, that said, I think you can leave out the previous stages too.

python-doctest/greet.py
```diff
@@ -0,0 +1,36 @@
+# def greet(name="World"):
```
iansedano, line 1: Again, some comments would be nice to indicate what the commented out versions mean.

python-doctest/identity.py
```diff
@@ -0,0 +1,7 @@
+def get_id(obj):
```
iansedano, line 1: I'd add a comment to just say "always fails"

python-doctest/context.py
```diff
@@ -0,0 +1,36 @@
+# context.py
```
iansedano, line 1: We don't need the file name for files in the materials repo.

python-doctest/options.txt
```diff
@@ -0,0 +1,10 @@
+# options.txt
```
iansedano, line 1: We don't need the filename in the materials repo

python-doctest/options1.txt
```diff
@@ -0,0 +1,10 @@
+>>> 5 < 7  # doctest: +DONT_ACCEPT_TRUE_FOR_1
```
iansedano, line 1: Maybe it makes sense to call this file something more descriptive, like "options_with_directives.txt" or something like that, likewise, the previous options.txt, could be called "options_no_directives.txt".

python-doctest/run_file.py
```diff
@@ -0,0 +1,3 @@
+import doctest
+
+doctest.testfile("tests_file.txt", verbose=True)
```
iansedano, line 3: This fails because `tests_file.txt` doesn't exist here

iansedano (changes requested):
Hi @lpozo, this is looking really nice. I've added a few comments to make the materials stand on their own a bit better.

I think we can get away with having slightly different names for some variations of files, but i'd also do a final check to make sure that the names of the files agree with the latest changes of the article.

lpozo:
@iansedano I did the update. Thanks for your feedback! Please, let me know if there is anything else I can do.

PR #335, opened 2022-11-21
Removing large desktop example and adding note about included script

iansedano:
Update to the materials for "How to Get a List of All Files in a Directory With Python"

- Removed "desktop large" which was an example of a directory tree used in a now deprecated example
- Removed "large_dir" and replaced with a script to generate it instead
- Adjusted the README

### DIFF
--- a/python-get-all-files-in-directory/create_large_dir.py
+++ b/python-get-all-files-in-directory/create_large_dir.py
@@ -4,6 +4,11 @@ from dataclasses import dataclass
 from pathlib import Path
 
 
+NUM_JUNK_FILES_LARGE = 300
+NUM_JUNK_FILES_MEDIUM = 25
+NUM_JUNK_FILES_SMALL = 5
+
+
 @dataclass
 class Item:
     name: str
@@ -22,8 +27,8 @@ folder_structure = Item(
                     [
                         Item(
                             "temp",
-                            [Item("2", junk_files=300)],
-                            junk_files=300,
+                            [Item("2", junk_files=NUM_JUNK_FILES_LARGE)],
+                            junk_files=NUM_JUNK_FILES_LARGE,
                         ),
                         Item("0.txt"),
                         Item("find_me.txt"),
@@ -35,10 +40,10 @@ folder_structure = Item(
                         Item(
                             "temporary_files",
                             [
-                                Item("logs", junk_files=300),
-                                Item("temp", junk_files=300),
+                                Item("logs", junk_files=NUM_JUNK_FILES_LARGE),
+                                Item("temp", junk_files=NUM_JUNK_FILES_LARGE),
                             ],
-                            junk_files=25,
+                            junk_files=NUM_JUNK_FILES_MEDIUM,
                         ),
                         Item("33.txt"),
                         Item("34.txt"),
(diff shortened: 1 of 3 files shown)


python-get-all-files-in-directory/create_large_dir.py
```diff
@@ -0,0 +1,83 @@
...
+            ],
+            junk_files=NUM_JUNK_FILES_SMALL,
+        ),
+        Item("temp", junk_files=NUM_JUNK_FILES_LARGE),
+        Item("temporary_files", junk_files=NUM_JUNK_FILES_LARGE),
+    ],
+)
+
+
+def create_item(item: Item, path_to: Path = Path("")):
```
gahjelle, line 64: We can explicitly use `Path.cwd()` for the current working directory:

```suggestion
def create_item(item: Item, path_to: Path = Path.cwd()) -> None:
```

We should probably also include the return annotation if we're gonna annotate the function.

python-get-all-files-in-directory/create_large_dir.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if item.children is None and item.junk_files is None:
+        path_to.joinpath(item.name).touch()
+        return
+
+    if item.children or item.junk_files:
+        root = path_to.joinpath(item.name)
+        root.mkdir()
+
```
gahjelle, line 73: Since we return in the previous `if` block, we know that either `.children` or `.junk_files` has some value. We should be able to skip this `if`:

```suggestion
    root = path_to.joinpath(item.name)
    root.mkdir()

```

python-get-all-files-in-directory/create_large_dir.py
```diff
@@ -0,0 +1,83 @@
...
+NUM_JUNK_FILES_LARGE = 300
+NUM_JUNK_FILES_MEDIUM = 25
+NUM_JUNK_FILES_SMALL = 5
+
+
+@dataclass
+class Item:
+    name: str
+    children: list[Item] = None
+    junk_files: int = None
```
gahjelle, line 16: Should we rename this to `num_junk_files` for explicitness and consistency?

Also, isn't `0` a better default value? If we change this, we should update the test later to be `if ... not item.junk_files` instead of `if ... item.junk_files is None`. Additionally, we don't need the `if item.junk_files` test since we'll only be looping over an empty range.

python-get-all-files-in-directory/create_large_dir.py
```diff
@@ -0,0 +1,83 @@
...
+
+NUM_JUNK_FILES_LARGE = 300
+NUM_JUNK_FILES_MEDIUM = 25
+NUM_JUNK_FILES_SMALL = 5
+
+
+@dataclass
+class Item:
+    name: str
+    children: list[Item] = None
```
gahjelle, line 15: We can probably use an empty list as a default value here. No need to fight with a `None` value later.

The only complication is that we'll need to import `field` from `dataclass` and set `children: list[Item] = field(default_factory=list)`. But then we can use `if not item.children` in the first `if` test in `create_items()` and skip `if item.children` completely.

gahjelle (approved):
Thanks for cleaning this up @iansedano 

I've added a few line comments on `create_large_dir.py`. These are just small quibbles to streamline the code ever so slightly.

PR #358, opened 2023-01-23
Add materials for Print Flush tutorial

martin-martin:
Co-authored-by: Ian <iansedano@gmail.com>
Co-authored-by: Jim Anderson <jima80525@gmail.com>

**Where to put new files:**

4. 🎉


flush-print/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+If you want to, you could even drill down further with some manual binary search to find the exact byte when the break first occurs.
+
+The idea of this script is that each of these printed numbers take up exactly 6 bytes in the buffer, which you assure with the format specifier in the f-string that you pass to `print()`. When the buffer gets filled up so much that it needs to flush during the execution of your script, then you run into the break in the printout, caused by the call to `time.sleep()` executing. Finally, when the script finishes execution after the break, Python flushes the rest of the buffer contents, which prints the remaining numbers to your terminal.
+
+Finally, you need to subtract from the current value of `SLIGHTLY_TOO_LARGE_FOR_BUFFER` because we assume that the buffer flushes continuously when it gets full, which means that whatever is in the buffer when the script is done should represent a full buffer.
+
+If you have better and more precise ways to calculate the buffer size, or have spotted a logical or computational mistake in these calculations, then please let us know!
```
iansedano, line 56: I think you've already said this above, so I'd take out this final paragraph.

flush-print/README.md
```diff
@@ -0,0 +1,64 @@
...
+This script gives you a way to approximate the buffer size that Python uses when writing to stdout on your operating system and configuration. Note that this can be quite different between different setups.
+
+The approach to calculate it is somewhat manual:
+
+1. Run the script
+1. Note if the numbers that it prints pause in between before the script finishes execution
+  - If they pause, remember the last number that got printed before the pause
+  - If they don't pause, increase the value of `SLIGHTLY_TOO_LARGE_FOR_BUFFER` by `10_000` and start from the top
+
+Continue doing this until your script paused when printing the numbers, and you noted the number that displayed last during the pause. Once you have that number, subtract it from the value that `SLIGHTLY_TOO_LARGE_FOR_BUFFER` currently has to calculate your buffer size estimation, for example:
```
iansedano, line 37: Maybe we include the "manual" nature of this operation to be 100% crystal clear:

```suggestion
Continue doing this until your script paused when printing the numbers, and you noted the number that displayed last during the pause. Once you have that number, manually subtract it from the value that `SLIGHTLY_TOO_LARGE_FOR_BUFFER` currently has to calculate your buffer size estimation, for example:
```

flush-print/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+bufsize = 80_000 - 10919
+print(bufsize) # 69081 <-- Your buffer size approximation
+```
+
+You can divide the number you get by `1000` to get an estimation of your buffer size for stdout in Kilobytes. In the example above, on a macOS system with a M1 chip, the buffer size of stdout when interacting with it through Python's `print()` would therefore be approximately 69 Kilobytes.
+
+If you want to, you could even drill down further with some manual binary search to find the exact byte when the break first occurs.
+
+The idea of this script is that each of these printed numbers take up exactly 6 bytes in the buffer, which you assure with the format specifier in the f-string that you pass to `print()`. When the buffer gets filled up so much that it needs to flush during the execution of your script, then you run into the break in the printout, caused by the call to `time.sleep()` executing. Finally, when the script finishes execution after the break, Python flushes the rest of the buffer contents, which prints the remaining numbers to your terminal.
```
iansedano, line 52: Adding encoding disclaimer and clarifying a couple things

```suggestion
The idea of this script is that each of these printed numbers take up exactly 6 bytes in the buffer. This is assuming that you're using an encoding in which numbers and spaces take up one byte. If so, then you can make sure each length is 6 characters with the format specifier in the f-string that you pass to `print()`.

At some point the buffer gets filled up so much that it needs to flush during the execution of your script. After it's flushed, it'll begin to fill up again. Ideally before it fills up again, you run into the call to `time.sleep()` executing. Finally, when the script finishes execution after the sleep, Python flushes the rest of the buffer contents, which prints the remaining numbers to your terminal.
```

flush-print/countdown_stderr.py
```diff
@@ -0,0 +1,8 @@
+import sys
+from time import sleep
+
```
iansedano, line 3: Nice example :)

flush-print/progress.py
```diff
@@ -0,0 +1,21 @@
...
+    print(
+        "\r[",
+        "#" * left,
+        " " * right,
+        "]",
+        f" {percent:.0f}%",
+        sep="",
+        end="",
+        flush=True,
+    )
```
iansedano, line 16: Nice example!

Feel free to disregard the following suggestion. I find something like this a bit more readable:

```suggestion
def progress(percent=0, width=30):
    progress = (width * percent // 100)
    bar = "#" * progress
    pad = " " * (width - progress)
    print(f"\r[{bar}{pad}]{percent:.0f}%", end="", flush=True)
```

flush-print/countdown_sepnewline.py
```diff
@@ -0,0 +1,6 @@
+from time import sleep
```
iansedano, line 1: Not 100% convinced by this example. Seems like a bit of a stretch and not what `sep` is for.

iansedano (changes requested):
Hey @martin-martin! Nice snippets!

I've left a few comments, let me know what you think :)

flush-print/progress.py
martin-martin: pretty rewrite! i had just taken this from bartosz print mega tutorial, but i like your code more. updated ✅

flush-print/countdown_sepnewline.py
martin-martin: Added an s for seconds and i'll update it with a note in the tutorial text as well.

flush-print/progress.py
martin-martin: Just added a space to make it look exactly the same as the other version

KateFinegan (approved):

PR #369, opened 2023-03-02
Update project code

martin-martin:
see comment below


martin-martin:
Hi @kylestratis I've applied a couple of updates to your source code, and also set up the project with the individual step folders that we talked about at some point.

I'll attempt to list all the update suggestions for you to be aware of and consider here, but you can also check out the `source_code_final/` folder, and especially [`detector.py`](https://github.com/realpython/materials/pull/369/files#diff-5d07ed66d0e72614299009e916af492947345b5fd4d169dee07bc842663b7270), even though there are also some other changes to the directory structure etc.

## ✅ Applied Changes

Here's a list of the notable changes, please ask back if you have questions or don't want to apply some of them:

- **Removed `eval_img/`**: The folder is unnecessary and not really used in your original code. I found the name not super descriptive, and because it wasn't in use, it seemed easier to keep an example test image directly in the source folder. This also makes calling the CLI tool with `--test -f` shorter, because it cuts the folder name from the path that users need to pass.

- **Added `.absolute()` in `validate()`:** I decided to add a call to `.absolute()` to `str(filepath)` [in line 116](https://github.com/realpython/materials/blob/5db67229be95d526ebc9e77b6466aa79ec8163b5/face-recognition/source_code_final/detector.py#L116) to reduce potential errors arising from relative paths:

```python
>>> form pathlib import Path
>>> p = Path("validation")

>>> p
PosixPath('validation')

>>> str(p)
'validation'

>>> str(p.absolute())
'/Users/martin/Documents/realpython/materials/face-recognition/source_code_step_4/validation'
```

- **Removed code comments:** Because we're writing tutorials, not just code, we aim to remove most code comments from within code blocks and use the surrounding text to explain the code, if helpful. I therefore removed most code comments from the code blocks, and threaded the information you had in there into the surrounding paragraphs.

- **Denotate a throwaway variable:** There was an unused variable assigned from tuple unpacking with `.textsize()` in [line 96](https://github.com/realpython/materials/blob/5db67229be95d526ebc9e77b6466aa79ec8163b5/face-recognition/source_code_final/detector.py#L96), which I replaced with an underscore. However, please see the requested changes below, because `.textsize()` is deprecated.

## ⛔️ Requested Changes

- **Add docstrings:** Please add docstrings to your functions.

- **Replace `.textsize()`:** [The method is deprecated](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.ImageDraw.textsize) and will be removed soon, so please update this part with one of its suggested alternatives.

## ✨ Additional Suggested Changes

I aimed to keep most of your source code without making any bigger changes. However, there are some suggestions I have that I think would improve the code and that I'd like you to consider. You don't have to implement these, but if you agree that they'd improve the project, then it could be nice to include them as well:

- **Auto-create necessary folders:** You're hard-coding the folder names of all three necessary directories right in your source code. In my opinion it would therefore be a good move to assure that these folders get created automatically, if they don't exist yet. You already use `pathlib` in your project, so you could quickly add this functionality at the top of your script:

```python
pathlib.Path("training").mkdir(exist_ok=True)
pathlib.Path("output").mkdir(exist_ok=True)
pathlib.Path("validation").mkdir(exist_ok=True)
```

- **Use `.rglob("*")`:** Instead of `.glob("*/*")` you could use [`.rglob("*")`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob), which is in my opinion more descriptive. But it's a matter of taste so feel free to ignore this suggestion.

- **Remove magic numbers:** When you draw the bounding boxes starting [in line 95](https://github.com/realpython/materials/blob/5db67229be95d526ebc9e77b6466aa79ec8163b5/face-recognition/source_code_final/detector.py#L95), you use a couple of magic numbers for colors. I'd suggest to define them as constants and then use the color names here in the code.

## ✨ Create a separate display function

This is a suggestion, but because it's quite a big change, it gets its own heading. The biggest one for the end!

Like I mentioned also in the other PR on the tutorial text, one of the most important aspects of the step-by-step project idea is that readers will be able to run the code and see something new happen at the end of each step.

Before my update pass, your steps didn't necessarily follow that idea. Step 3 is especially tricky in this sense, because it ends without any clear output in the code. I've fixed that by printing out the predictions, then later removing the print call.

However, in my opinion it would be better if you'd rewrite `recognize_faces()` in a way so that you split the function into two separate functions. It's currently quite long, which shows in multiple ways:

- The Preflight complains about this section length
- The function code is long
- You can't easily end after step 3 with a clear output for the readers
- It does two different, somewhat unrelated, things

Speaking from the perspective of writing clean code:

>Functions should do one thing. They should do it well. They should do it only. ([Source](https://www.goodreads.com/quotes/9395684-functions-should-do-one-thing-they-should-do-it-well))

I think that if you split up `recognize_faces()` into `recognize_faces()` and `display_predictions()` (or something similar) that would solve all the issues mentioned above in one go.

So my suggestion is to take out all the PIL related code into a separate function, that you create and describe in Step 4. The function takes the output of `recognize_faces()` as its input and doesn't need to be callable through the CLI. This is definitely a bigger shuffle, however, so I wouldn't want to take a decision like that without you as the author, but I think it would improve the source code as well as the tutorial overall.

What do you think?

kylestratis:
@martin-martin Thank you for the suggestions, and I'm working on implementing them. As for a separate display function, I'm not sure this will work to make things less confusing (or at least more readable).

The problem here is that in order for multiple face detection to work properly (or, rather, its display), the Pillow Draw object needs to be both set up, deleted, and displayed outside of the for loop of `recognize_faces()`, while the actual drawing is done within. Besides departing from the Python ideal of flat being better than nested (which is in a wonderful tension with DRY and the single responsibility principle), making a separate function work would still require PIL code to be in `recognize_faces()`, and the drawing portion of the code will require at least 6 parameters (the 4 bounding box coordinates, the draw object, and the result). 

I'm still willing to make the change, however I wanted to flag this before starting the work in case it hinders the code comprehensibility.

martin-martin:
Hi @kylestratis . I'll go ahead and fix that misspelled Affleck bro--not sure how that sneaked in but thanks for noticing :)

I don't quite know what you mean with that the validation image got removed. The `validation/` folder still has a bunch of images in there. I've removed images from the steps where they aren't necessary to keep the overall file size of the repository a bit in check. That's why sometimes the `training/` and `validation/` folders are empty, and sometimes not.

I did a quick check and couldn't identify what unknown image you noticed got added to the `training/` directory, but I'll let you take care of that, thanks!

---

Regarding the suggested code change to separate recognizing and drawing faces more clearly, I've discussed the situation also with @gahjelle and he suggested to refactor the code with two helper functions:

1. `_recognize_face()`
2. `_display_face()`

This way you could reduce the amount of code that's nested under the `for` loop significantly and make it overall easier to follow the code logic. It would also help towards keeping the code that you add in Step 3 focused on _recognizing_ faces, and the code you add in Step 4 focused on _displaying_ the results.

Making this conceptual separation happen is a requirement for the tutorial, how exactly you go about it is your choice. However, below is the suggestion that @gahjelle came up with (slightly modified, contextualized, and tested by me) based on your code. You're welcome to use it if you want.

In _Step 3: Recognizing Unlabeled Faces_ you could build out the relevant functions like so:

```python
from collections import Counter

# ...

def recognize_faces(
    image_location: str,
    model: str = "hog",
    encodings_location: str = DEFAULT_ENCODINGS_PATH,
) -> None:
    with open(encodings_location, "rb") as f:
        loaded_encodings = pickle.load(f)

    input_image = face_recognition.load_image_file(image_location)
    input_face_locations = face_recognition.face_locations(
        input_image, model=model
    )
    input_face_encodings = face_recognition.face_encodings(
        input_image, input_face_locations
    )

    for bounding_box, unknown_encoding in zip(
        input_face_locations, input_face_encodings
    ):
        name = _recognize_face(unknown_encoding, loaded_encodings)
        print(name, bounding_box)


def _recognize_face(unknown_encoding, loaded_encodings):
    boolean_matches = face_recognition.compare_faces(
        loaded_encodings["encodings"], unknown_encoding
    )
    votes = Counter(
        name
        for match, name in zip(boolean_matches, loaded_encodings["names"])
        if match
    )
    if votes:
        return votes.most_common(1)[0][0]
```

That way you'd skip adding anything PIL related while you're still in Step 3, and you're splitting the functionality between the main function `recognize_faces()`, and a helper function, `_recognize_face()`.

Note that you could also skip printing `bounding_box`, which you'll only use in the next step, but I added it because it seemed to me that it could be a nice segue into the next step.

Then, in _Step 4: Displaying Results_, you can go back into `recognize_faces()` to add the PIL related code, as well as another helper function, `_display_face()`, that would handle creating the visuals:

```python
from PIL import Image, ImageDraw

# ...

def recognize_faces(
    image_location: str,
    model: str = "hog",
    encodings_location: str = DEFAULT_ENCODINGS_PATH,
) -> None:

    # ...

    pillow_image = Image.fromarray(input_image)
    draw = ImageDraw.Draw(pillow_image)

    for bounding_box, unknown_encoding in zip(
        input_face_locations, input_face_encodings
    ):
        name = _recognize_face(unknown_encoding, loaded_encodings)
        # print(name, bounding_box)  # Removed
        if name:
            _display_face(draw, bounding_box, name)

    del draw
    pillow_image.show()


def _recognize_face(unknown_encoding, loaded_encodings):
    # ...


def _display_face(draw, bounding_box, name):
    top, right, bottom, left = bounding_box
    draw.rectangle(((left, top), (right, bottom)), outline=(51, 51, 255))
    _, caption_height = draw.textsize(name)
    draw.rectangle(
        ((left, bottom - caption_height - 10), (right, bottom)),
        fill=(51, 51, 255),
        outline=(51, 51, 255),
    )
    draw.text(
        (left + 6, bottom - caption_height - 5),
        name,
        fill=(255, 255, 255, 255),
    )
```

That way you're keeping all the display-related code in Step 4 and reduce the complexity of your `for` loop, which will make it better understandable for your readers.

---

While we understand that David reviewed your project code when you were first working on this tutorial, and I've tried to keep that in mind during my review, we thought it would still be worth the additional effort in this case to make the code more readable. We want the tutorial to stick around for a long time and keep providing value to learners after all :)

Hope that all makes sense and still fits into your weekend plan! LMK if you want to discuss any of the updates, best probably on Slack. Thanks for your continued work on this tutorial!

kylestratis:
Ah that's a great suggestion and looks like it should work! I'll work on implementing it and getting the rest of the suggestions in over the weekend. Thank you for all the help! I agree with tightening this up as much as possible, I think it will continue to serve as a great introduction to computer vision for lots of people.

kylestratis:
Hi @martin-martin - I've made these changes and pushed them. I also updated the "unknown" image in later steps with one face that should be recognized and one that wouldn't be, as suggested. As for the image added to training - it turns out this was an artifact of how Github displays file changes. It was the unknown image added to the source_code_final directory, but with the training subtree open in Github's changed file viewer, it appears to be at or near the same level as the contents of the training directory.

PR #372, opened 2023-03-11
Source code for python-input-integer

profcalculus:
Source files for the article [How to Read Python Input As Integers](https://realpython.com/python-input-integers).


profcalculus:
Hi @KateFinegan,
I've added the code examples as promised. 
Probably I missed something in the docs, but I couldn't get the `black` and `flake8` defaults to play together nicely - I had to specify `black -l79` to get black's output past flake8. (?)
Anyway, let me know what you think.

gahjelle:
@profcalculus 

In https://github.com/realpython/materials/blob/master/pyproject.toml we're setting the line length for `black` to be 79. I'm not sure why that wasn't picked up for you automatically. Anyway, doing `-l 79` should do the same thing :+1:

gahjelle (approved):
Great, thanks for adding these, @profcalculus 

I've updated the PR with a few minor changes, mainly changing comments into proper docstrings for the functions.

PR #380, opened 2023-04-26
Sample code for the article on global variables in functions

lpozo:
**Where to put new files:**

4. 🎉


python-use-global-variable-in-function/conf.py
```diff
@@ -0,0 +1,13 @@
+config = {
+    "api_key": None,
```
gahjelle, line 2: Should we remove `api_key` here as well, for consistency?

python-use-global-variable-in-function/nasa.py
```diff
@@ -0,0 +1,31 @@
+import webbrowser
```
gahjelle, line 1: This file is named `nasa_api.py` in the tutorial.

gahjelle (commented):
The materials look good @lpozo 

I only had two small comments.

PR #393, opened 2023-06-01
Sample code for the article on flattening list

lpozo:
**Where to put new files:**

4. 🎉


gahjelle (commented):
We can probably remove all the individual `flatten_*` files, maybe except for the NumPy example.

Apart from that, this looks good to me.

python-flatten-list/flatten_numpy_flatten.py
```diff
@@ -0,0 +1,6 @@
+import numpy as np
+
+matrix = [list(range(100))] * 4
+arr = np.array(matrix)
```
gahjelle, line 4: This ends up being slightly wrong since we flatten `matrix` while the NumPy array is named `arr`.

NumPy has its own `np.arange()` so it feels a bit backwards to use `range()` here and then convert to a NumPy array later. I'd suggest either using the literal matrix from the article, or doing something like this instead:

```suggestion
matrix = np.arange(400).reshape(100, 4)
```

gahjelle (commented):
Thanks for the update @lpozo 

There is a small issue with the NumPy example. Otherwise, this looks good to me!

gahjelle (approved):
:+1:

PR #420, opened 2023-08-02
Python polars article code

hfhoffman1144:
Python Polars showcase code


python-polars/README.md
```diff
@@ -0,0 +1,3 @@
+# Pandas-Groupby
```
KateFinegan, line 1: @hfhoffman1144 Is there a reason that Pandas-Groupby is the title here? I would suggest changing it to match the tutorial's title.

python-polars/README.md
```diff
@@ -0,0 +1,3 @@
...
+
+Supporting code for the article [Python Polars: A Lightening-Fast DataFrame Library](https://realpython.com/polars-python-lightening-fast-dataframe-library/).
```
gahjelle, line 3: I suggest that we add a small paragraph about necessary dependencies as well. I think the readers need to do something like `python -m pip install "polars[numpy,pandas]" requests matplotlib` in order to run the examples.

python-polars/downloads.py
```diff
@@ -0,0 +1,14 @@
+import requests
+import pathlib
+
+
+def download_file(file_url: str, local_file_path: pathlib.Path) -> None:
+    "Download a file from the URL and save it with the specified file name."
```
gahjelle, line 6: Per convention, we should use a triple-quoted string for doc-strings:

```suggestion
    """Download a file from the URL and save it with the specified file name."""
```

python-polars/downloads.py
```diff
@@ -0,0 +1,14 @@
+import requests
+import pathlib
```
gahjelle, line 2: Minor quibble, but we should put the standard library on top of the imports:

```suggestion
import pathlib
import requests
```

python-polars/scanning_data.py
```diff
@@ -0,0 +1,31 @@
+import pathlib
+from downloads import download_file
+import polars as pl
```
gahjelle, line 3: Another small quibble. Let's do the local import after the third-party import:

```suggestion
import pathlib
import polars as pl
from downloads import download_file
```

gahjelle (commented):
Thanks for putting together this opt-in, @hfhoffman1144 .

I only have a few quick comments that you can find in the line comments.

python-polars/README.md
hfhoffman1144: Sorry about that, I copied the README template from another article and forgot to change the main header 😅

gahjelle (approved):
Looks great! :+1:

PR #429, opened 2023-08-25
Multiple Exceptions

eyrei123:
**Where to put new files:**

4. 🎉


gahjelle (commented):
Hey @eyrei123 

Thanks for preparing these materials. I think these look great and I only have three small requests before approving:

1. We usually don't include the comment with the file name in the materials repo, since the file name is already in the file :blush: Can you please remove all of those?
2. The name of `more_exception_plusKeyBoardInterrupt_suppress.py` is non-standard with the sudden CamelCase and a bit convoluted. How about naming it `more_exception_suppress_keyboardinterrupt.py` or something like that instead?
3. Also, the `more_`-names are not great. Can we be a bit more concrete/descriptive in those file names? For example, the suppress variants could be named something like `archiver.py`

Thanks for putting these together!

gahjelle (approved):
Looks great!

PR #435, opened 2023-09-12
Sample code for the article on f-strings

lpozo:
**Where to put new files:**

4. 🎉


python-f-string/basic_f_strings.py
```diff
@@ -0,0 +1,5 @@
+name = "Jane"
+age = 25
+f"Hello, {name}! You're {age} years old."
+
+f"{2 * 21}"
```
martin-martin, line 5: I think that for providing these code blocks as files it makes sense to wrap REPL session output into calls to `print()`. That's at least what I've been doing, to make it convenient for readers to run.

Do we have a guideline on doing/not doing that @gahjelle ?

lpozo: I guess that wrapping the code in calls to `print()` can confuse the reader because the code differs from the code in the article.

martin-martin (approved):

PR #483, opened 2024-01-05
Materials for Range tutorial

gahjelle:
**Where to put new files:**

4. 🎉

### DIFF
--- a/python-range/float_range.py
+++ b/python-range/float_range.py
@@ -1,11 +1,13 @@
-import math
 from dataclasses import dataclass, field
+from math import ceil, isclose
 
 
 @dataclass
 class FloatRange:
+    """Range of numbers that allows floating point numbers."""
+
     start: float | int
-    stop: float | int = None
+    stop: float | int | None = None
     step: float | int = 1.0
 
     def __post_init__(self):
@@ -20,7 +22,7 @@ class FloatRange:
             raise ValueError("'start' must be a floating point number")
         if not isinstance(self.stop, float | int):
             raise ValueError("'stop' must be a floating point number")
-        if not isinstance(self.step, float | int) or self.step == 0:
+        if not isinstance(self.step, float | int) or isclose(self.step, 0):
             raise ValueError("'step' must be a non-zero floating point number")
 
     def __iter__(self):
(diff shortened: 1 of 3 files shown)


python-range/float_range.py
```diff
@@ -0,0 +1,97 @@
+import math
+from dataclasses import dataclass, field
+
+
+@dataclass
+class FloatRange:
+    start: float | int
+    stop: float | int = None
```
bzaczynski, line 8: This is technically `float | int | None`.

python-range/float_range.py
```diff
@@ -0,0 +1,97 @@
...
+        return round((element - self.start) / self.step)
+
+
+@dataclass
+class _FloatRangeIterator:
+    start: int
+    stop: int
+    step: int
+    _num_steps: int = field(default=0, init=False)
+
```
bzaczynski, line 87: The iterator protocol requires that an iterator object implements the `.__iter__()` method as well:

```suggestion
    def __iter__(self):
        return self
```

Without it, you won't be able to loop over the iterator itself or pass it into a `list()`:

```python
>>> for value in _FloatRangeIterator(1, 3, 0.5):
...     print(value)
...
1.0
1.5
2.0
2.5

>>> it = _FloatRangeIterator(1, 3, 0.5)
>>> list(it)
[1.0, 1.5, 2.0, 2.5]
```

python-range/float_range.py
```diff
@@ -0,0 +1,97 @@
...
+    def __next__(self):
+        """Calculate the next element in the iteration."""
+        element = self.start + self._num_steps * self.step
+        if any(
+            self.step > 0 and element >= self.stop,
+            self.step < 0 and element <= self.stop,
```
bzaczynski, line 93: When `.step` is equal to zero, the iteration will never stop. The original `range()` function rejects zero as the step, so you could perhaps add similar validation in the `.__post_init__()` method.

python-range/float_range.py
```diff
@@ -0,0 +1,97 @@
...
+    stop: float | int = None
+    step: float | int = 1.0
+
+    def __post_init__(self):
+        """Validate parameters."""
+        # Only one argument is given
+        if self.stop is None:
+            self.stop = self.start
+            self.start = 0
+
```
bzaczynski, line 17: If step is zero, we should raise an error to handle edge cases:

```python
if math.isclose(self.step, 0):
    raise ValueError("'step' must not be zero")
```

bzaczynski (changes requested):
@gahjelle Overall, your code looks good to me. I only flagged a few minor edge cases that it would be nice to handle. Let me know what you think. Thanks!

python-range/float_range.py
gahjelle: This is checked by `if not isinstance(self.step, float | int) or self.step == 0` below. I'm switching to `isclose()`.

python-range/float_range.py
gahjelle: I've emphasized in the docstring that `_FloatRangeIterator` is non-public and should only be instantiated through `FloatRange` which does the validation. I'll leave the foot-gun for those desperate to use it ...

gahjelle:
Thanks @bzaczynski for the detailed review. I've updated the code.

bzaczynski (approved):
@gahjelle Thank you for the updates. They looks good to me :+1:

PR #555, opened 2024-07-10
dircheck.py: fix folder naming violations not causing exit code 1

dbader:


dbader:
@gahjelle I renamed `pivot_tables` to `pandas-pivot-table` to be a closer match to the article slug and resolve the dircheck naming error.

gahjelle (approved):
Looks great to me!

I can coordinate with @brendaweles so that we fix the opt-in link as we merge.

PR #561, opened 2024-07-29
Add materials for functional programming update

martin-martin:
**Where to put new files:**

4. 🎉


martin-martin:
ah, sorry for the commit name, I thought I had done one before adding those flags, but now it's all in one :P

I've just added the code examples, split them into separate files, and added comments to make it somewhat useful. Was maybe more work than it's worth, considering that this isn't super useful material anyway 🤔

Maybe we should eventually replace it with some actually useful examples that go beyond these toy code snippets.

But maybe also ok now, so that we have a materials opt in we can add. LMK.

brendaweles (approved):

brendaweles (approved):

PR #567, opened 2024-08-14
Sample code for the article on lists vs tuples

lpozo:
**Where to put new files:**

4. 🎉


python-lists-tuples/unpacking.py
```diff
@@ -0,0 +1,20 @@
...
+# Using a temporary variable
+temp = a
+a = b
+b = temp
+(a, b)
+a = "foo"
+b = "bar"
+# Using unpacking
+a, b = b, a
+a, b
```
martin-martin, line 20:
```suggestion
a = "foo"
b = "bar"
# Using a temporary variable
temp = a
a = b
b = temp
print(a, b)

a = "foo"
b = "bar"
# Using unpacking
a, b = b, a
print(a, b)
```

martin-martin (changes requested):
@lpozo I'd suggest to use underscores for the script names, so that they can potentially also be moduels. Probably no one will use them like that, but I think we're generally naming Python files with underscores in this repo.

Also, a small change to show more output in the unpacking file.

martin-martin (approved):

martin-martin:
Great, looks good to me ✅

brendaweles (approved):

PR #581, opened 2024-09-13
Sample code for the article on closures

lpozo:
**Where to put new files:**

4. 🎉


python-closure/closure.py
```diff
@@ -0,0 +1,16 @@
...
+
+#     def inner_func():
+#         print(f"Hello, {name}!")
+
+#     return inner_func
+
+
+def outer_func():
+    name = "Pythonista"
+    return lambda name=name: print(f"Hello, {name}!")
```
bzaczynski, line 12: :warning: **Fact Check**

@lpozo Actually, I just realized this is a wrong syntax for a lambda expression. This is the correct one:

```suggestion
    return lambda name: print(f"Hello, {name}!")
```

Unfortunately, this mistake is also present in the tutorial text.

bzaczynski (changes requested):
@lpozo There's one small detail to fix in the lambda expression example (same in the tutorial text.)

python-closure/closure.py
lpozo: It doesn't work if we don't set the argument:

```python
>>> def outer_func():
...     name = "Pythonista"
...     return lambda name: print(f"Hello, {name}!")
...     
... 
>>> func = outer_func()
>>> func()
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    func()
TypeError: outer_func.<locals>.<lambda>() missing 1 required positional argument: 'func'
```

lpozo: The correct syntax is:

```python
>>> def outer_func():
...     name = "Pythonista"
...     return lambda: print(f"Hello, {name}!")
...     
... 
>>> func = outer_func()
>>> func()
Hello, Pythonista!
```

bzaczynski (approved):

brendaweles (approved):

PR #587, opened 2024-09-26
Add examples from Python 3.13: Cool New Features

gahjelle:
**Where to put new files:**

4. 🎉

### DIFF
--- a/python-313/typing/deprecations.py
+++ b/python-313/typing/deprecations.py
@@ -1,0 +1,79 @@
+"""Demonstration of PEP 702: Marking deprecations using the type system
+
+Use PyLance in VS Code by setting Python › Analysis: Type Checking Mode or run
+the Pyright CLI:
+
+    $ python -m pip install pyright $ pyright --pythonversion 3.13 .
+
+Note that showing warnings requires setting the reportDeprecated option in
+Pyright. This is done in pyproject.toml.
+"""
+
+from typing import overload
+from warnings import deprecated
+
+
+@deprecated("Use + instead of calling concatenate()")
+def concatenate(first: str, second: str) -> str:
+    return first + second
+
+
+@overload
+@deprecated("add() is only supported for floats")
+def add(x: int, y: int) -> int: ...
+@overload
+def add(x: float, y: float) -> float: ...
+
+
+def add(x, y):
+    return x + y
+
+
+class Version:
+    def __init__(self, major: int, minor: int = 0, patch: int = 0) -> None:
+        self.major = major
+        self.minor = minor
+        self.patch = patch
+
+    @property
+    @deprecated("Use .patch instead")
+    def bugfix(self):
+        return self.patch
+
+    def bump(self, part: str) -> None:
+        if part == "major":
+            self.major += 1
+            self.minor = 0
+            self.patch = 0
+        elif part == "minor":
+            self.minor += 1
+            self.patch = 0
+        elif part == "patch":
+            self.patch += 1
+        else:
+            raise ValueError("part must be 'major', 'minor', or 'patch'")
+
+    @deprecated("Use .bump() instead")
+    def increase(self, part: str) -> None:
+        return self.bump(part)
+
+    def __str__(self):
+        return f"{self.major}.{self.minor}.{self.patch}"
+
+
+@deprecated("Use Version instead")
+class VersionType:
+    def __init__(self, major: int, minor: int = 0, patch: int = 0) -> None:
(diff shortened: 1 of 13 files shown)


python-313/typing/readonly.py
```diff
@@ -0,0 +1,48 @@
+"""Demonstration of PEP 705: TypedDict: read-only items
+
+Use PyLance in VS Code by setting Python › Analysis: Type Checking Mode or run
+the Pyright CLI:
+
+    $ python -m pip install pyright $ pyright --pythonversion 3.13 .
```
bzaczynski, line 6: Should these two commands be on separate lines?

python-313/typing/readonly.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+# class PythonVersion(TypedDict):
+#     version: str
+#     release_year: int
+
+
+class Version(TypedDict):
+    version: ReadOnly[str]
+    release_year: ReadOnly[NotRequired[int | None]]
```
bzaczynski, line 26: Does the order matter, i.e., do you know if these two are equivalent?

```python
ReadOnly[NotRequired[int | None]]
NotRequired[ReadOnly[int | None]]
```

python-313/typing/readonly.py
```diff
@@ -0,0 +1,48 @@
...
+
+def get_version_info(ver: Version) -> str:
+    if "release_year" in ver:
+        return f"Version {ver['version']} released in {ver['release_year']}"
+    else:
+        return f"Version {ver['version']}"
+
+
+# Only allowed to use PythonVersion instead of Version if the fields are ReadOnly
+print(get_version_info(py313))
```
bzaczynski, line 48: I find this counter-intuitive since the `Version` and `PythonVersion` classes specify separate types that don't belong to the same type hierarchy. They share the same structure, so they're essentially the same type by means of static duck typing, but they don't explicitly define a protocol. I guess, both are dicts with the same set of key-value pairs, so that should be enough for the type checker to treat them as equals.

python-313/typing/deprecations.py
```diff
@@ -0,0 +1,79 @@
+"""Demonstration of PEP 702: Marking deprecations using the type system
+
+Use PyLance in VS Code by setting Python › Analysis: Type Checking Mode or run
+the Pyright CLI:
+
+    $ python -m pip install pyright $ pyright --pythonversion 3.13 .
```
bzaczynski, line 6: Should these two commands be on separate lines?

python-313/typing/deprecations.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+concatenate("three", "thirteen")
+add(3, 13)
+VersionType(3, 13)
+
+version = Version(3, 13)
+version.increase("patch")
+print(version)
+print(version.bugfix)
```
bzaczynski, line 79: PyCharm is smart enough to recognize the new decorator and uses strike-through formatting:

image

bzaczynski (approved):
@gahjelle Looks good to me. I left a few comments/questions, but otherwise, approved!

python-313/typing/readonly.py
gahjelle: Oops, yes, definitely. It seems that the automatic comment formatter I ran is a bit eager.

python-313/typing/deprecations.py
gahjelle: Great. Yes, that looks quite similar to VS Code. I'll add PyCharm to the top comment.

image

python-313/typing/readonly.py
gahjelle: Yes, according to the PEP, they can be nested in any order: https://peps.python.org/pep-0705/#interaction-with-other-special-types

This works for me in VS code/Pyright as well.

python-313/typing/readonly.py
gahjelle: Right, I believe that TypedDict is all about structural typing. The [PEP](https://peps.python.org/pep-0705/) talks a bit about how the capabilities of `ReadOnly` can be done with some complicated use of protocols instead.

I haven't quite internalized the rules for when one TypedDict can pass as the other, but everytime I sit down and think for a while, it seems to add up :see_no_evil:

gahjelle:
Thanks for the review @bzaczynski I've done a few updates to the extra typing files!

bzaczynski (approved):

brendaweles (approved):

PR #583, opened 2024-09-20
Add Code for Selenium Web Automation Tutorial Update

martin-martin:
**Where to put new files:**

- New files should go into a top-level subfolder, named after the article slug. For example: `my-awesome-article`

**How to merge your changes:** 

1. [Make sure the CI code style tests all pass (+ run the automatic code formatter if necessary).](https://github.com/realpython/materials/blob/master/README.md)
2. Find an RP Team member on Slack and ask them to review & approve your PR.
3. Once the PR has one positive ("approved") review, GitHub lets you merge the PR.
4. 🎉

PR #596, opened 2024-10-14
Add slimmed-down Selenium project code

martin-martin:
**Where to put new files:**

4. 🎉


martin-martin:
@bzaczynski this is my attempt to create a slimmed-down version of the project that primarily focuses on Selenium web interactions (now following POM), but also keep the basic functionality of the music player (and a reduced---but existing---separation of concerns).

I've tried to build the project in a way so that it can lead into the project that you wrote, with the idea in mind that learners could follow your SbSP as a next step if they're interested in the music player project. At the same time, I tried to keep it simple and reduced enough that learners who are only there for an intro to Selenium would get their money's worth.

Hope this works and looking forward to your thoughts!

Linking the other PR for reference: https://github.com/realpython/materials/pull/583

python-selenium/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+The only direct dependency for this project is [Selenium](https://selenium-python.readthedocs.io/). You should use a Python version of at least 3.8.
+
+## Run the Bandcamp Discover Player
+
+To run the music placer, navigate to the `src/` folder, then execute the module from your command-line:
+
+```sh
+(venv) $ cd src/
+(venv) $ python -m bandcamp
```
bzaczynski, line 23: :wrench: **Technical Recommendation**

That doesn't work:

```
(python-selenium) $ cd src/
(python-selenium) $ python -m bandcamp/
/home/bartek/.virtualenvs/python-selenium/bin/python: No module named bandcamp/
```

A better option would be to install the package into the venv first. That way, it won't matter where you'll run the module from:

```
(python-selenium) $ python -m pip install .
(python-selenium) $ python -m bandcamp/
```

bzaczynski (changes requested):
@martin-martin I left a few comments for you. Let me know what you think.

python-selenium/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Then, install the requirements:
+
+```sh
+(venv) $ python -m pip install -r requirements.txt
+```
+
+The only direct dependency for this project is [Selenium](https://selenium-python.readthedocs.io/). You should use a Python version of at least 3.10, which is necessary to support [structural pattern matching](https://realpython.com/structural-pattern-matching/).
+
+You'll need a [Firefox Selenium driver](https://selenium-python.readthedocs.io/installation.html#drivers) called `geckodriver` to run the project as-is. Make sure to [download and install](https://github.com/mozilla/geckodriver/releases) it before running the project.
```
bzaczynski, line 17: @martin-martin For some reason, I was having trouble starting the script due to problems with my Firefox browser. Your script would just hang without showing any output, leaving me a bit puzzled. However, I had a hunch where to look for the problem, so switching to Chrome fixed it for me. I'm wondering if we could give the users some heads up in the README about potential issues and how to troubleshoot them.

python-selenium/src/bandcamp/app/tui.py
```diff
@@ -0,0 +1,74 @@
...
+                    print("Exiting the player...")
+                    break
+                case _:
+                    print("Unknown command. Try again.")
+
+
+def play(player, track_number=None):
+    """Play a track and shows info about the track."""
+    player.play(track_number)
+    print(player._current_track._get_track_info())
```
bzaczynski, line 46: @martin-martin It would be nice to include some rudimentary error handling. For example, when I request to play a track outside the valid range, the script crashes abruptly.

python-selenium/src/bandcamp/web/element.py
```diff
@@ -0,0 +1,84 @@
...
+            artist=self._parent.find_element(*self.ARTIST).text,
+            genre=genre,
+            url=clean_url,
+        )
+
+    def _get_play_button(self):
+        return self._parent.find_element(*self.PLAY_BUTTON)
+
+
+class DiscoverTrackList(WebComponent, DiscoverPageLocator, TrackLocator):
```
bzaczynski, line 48: @martin-martin I believe I mentioned this before, but I'm not a big fan of abusing inheritance for code reuse like here, where you inherit the constans. Class mixins are primarily about combining _behaviors_. When you only have data attributes, it's clearer to use composition and delegation instead.

For example, instead of this:

```python
view_more_button = self._driver.find_element(*self.PAGINATION_BUTTON)
```

...you'd have that:

```python
class DiscoverTrackList(WebComponent):
    # ...
    view_more_button = self._driver.find_element(*DiscoverPageLocator.PAGINATION_BUTTON)
```

While more verbose, this version tells me exactly which web component the locator belongs to, so I don't have to hunt for it in the source code.

python-selenium/src/bandcamp/web/element.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def _get_available_tracks(self) -> list:
+        """Find all currently available tracks in the Discover section."""
+        self._wait.until(
+            lambda driver: any(
+                e.is_displayed() and e.text.strip()
+                for e in driver.find_elements(*self.ITEM)
+            ),
+            message="Timeout waiting for track text to load",
+        )
```
bzaczynski, line 71: @martin-martin This lambda expression is relatively dense, which is why I'd personally opt for a regular function with a descriptive name so that the code reads more easily:

```python
def _get_available_tracks(self) -> list:
    self._wait.until(track_text_loaded,
        message="Timeout waiting for track text to load",
    )

    # ...

def track_text_loaded(driver):
    return any(
        e.is_displayed() and e.text.strip()
        for e in driver.find_elements(*self.ITEM)
    )
```

bzaczynski (commented):
@martin-martin Hi Martin!

The code looks great overall. I only left a few minor comments for your consideration, but no explicit requests for change.

Take care!

brendaweles (approved):

PR #620, opened 2024-12-16
Sample code for the article on for loops

lpozo:
**Where to put new files:**

4. 🎉


python-for-loop/break_continue.py
```diff
@@ -0,0 +1,14 @@
...
+target = 5
+for number in numbers:
+    print(f"Processing {number}...")
+    if number == target:
+        print(f"Target found {target}!")
+        break
+
+numbers = [1, 2, 3, 4, 5, 6]
+for number in numbers:
+    print(f"{number = }")
```
martin-martin, line 11: flake8 compains and I was also wondering about it, I've generally only seen this used without whitespace:

```suggestion
    print(f"{number=}")
```

`pycodestyle`, which implements the check, seems to be stubborn about keeping it this way: https://github.com/PyCQA/pycodestyle/issues/1201

```sh
$ flake8 break_continue.py 
break_continue.py:11:20: E251 unexpected spaces around keyword / parameter equals
break_continue.py:11:22: E202 whitespace before '}'
break_continue.py:11:22: E251 unexpected spaces around keyword / parameter equals
```

But I know you're using space around the equals sign also [in the f-string tutorial](https://realpython.com/python-f-strings/#self-documenting-expressions-for-debugging).

So IDK, I guess the linters don't fail on it?

If you decide to change it, you'd also have to change it in the written tutorial.

python-for-loop/upacking.py
```diff
@@ -0,0 +1,4 @@
+points = [(1, 4), (3, 6), (7, 3)]
+
+for x, y in points:
+    print(f"{x = } and {y = }")
```
martin-martin, line 4: Same as in the previous comment, flake8 doesn't like the whitespace:

```suggestion
    print(f"{x=} and {y=}")
```

python-for-loop/upacking.py
```diff
@@ -0,0 +1,4 @@
+points = [(1, 4), (3, 6), (7, 3)]
```
martin-martin, line 1: You forgot an `n` in the file name:

upacking.py --> unpacking.py

python-for-loop/nemu.py
```diff
@@ -0,0 +1,7 @@
+def display_menu(options):
```
martin-martin, line 1: I assume the file name should be "menu.py"

martin-martin (changes requested):
LGTM, two file name typos and a question about whitespaces around the self-documenting expression in f-strings.

python-for-loop/break_continue.py
lpozo: If I remember correctly, in one of my previous articles I used this feature without spaces and I guess it was Geir Arne who suggested adding the spaces for readability. However, I don't think this is a written rule.

martin-martin: Ok, yeah I'm fine with leaving it considering it passes our linter checks. @gahjelle should we note this somewhere, that we'll ignore E251 and E202 for these cases?

martin-martin (approved):

brendaweles (approved):

Issue #621, opened 2024-12-17
python-get-all-files-in-directory: create_large_dir.py contains wrong variable name


GrippenDynamik:
**Describe the bug**
https://github.com/realpython/materials/tree/master/python-get-all-files-in-directory
create_large_dir.py 
line 69
"item.junk_files" should be "item.num_junk_files"

**To Reproduce**
1. Download https://github.com/realpython/materials/tree/master/python-get-all-files-in-directory
2. Run create_large_dir.py
3. AttributeError: 'Item' object has no attribute 'junk_files'. Did you mean: 'num_junk_files'?

**Expected behavior**
create_large_dir.py should run without errors.

**Desktop (please complete the following information):**
 - OS: Linux Mint 22 Cinnamon 64-bit
 - Python 3.12.3

dbader:
Thanks for the report @GrippenDynamik. Will be fixed in https://github.com/realpython/materials/issues/621

PR #622, opened 2024-12-18
Fix variable name and add `exist_ok=True`

lpozo:
Fixes https://github.com/realpython/materials/issues/621


martin-martin (approved):
LGTM and works ✅

PR #625, opened 2025-01-08
Django diary update

acsany:
Here are the updated files for my tutorial update. There are no relevant changes. Mostly updated links and reformatting.


gahjelle (approved):
Looks great!

PR #643, opened 2025-02-23
Materials for Python bytearray

bzaczynski:
**Where to put new files:**

4. 🎉


lpozo (changes requested):
@bzaczynski Hey!

I wonder whether we can add some short examples from all the REPL examples we have in the article. Maybe the different ways to create bytearray objects and other useful examples. Of course, we'd need to add print() calls to produce output because we're not in a REPL. What do you think? Maybe, we can group the examples in a couple of `.py` file.

brendaweles (approved):

lpozo (approved):

PR #693, opened 2025-08-12
Add Materials for "A Close Look at a FastAPI Example Application"

acsany:
Trello: https://trello.com/c/0QkmGHJA/1367-using-fastapi-to-build-python-web-apis-update


fastapi-python-web-apis/main.py
```diff
@@ -0,0 +1,174 @@
...
+    remaining_items_count: int
+
+
+@app.get("/", tags=["Random Playground"])
+async def home():
+    return {"message": "Welcome to the Randomizer API"}
+
+
+@app.get("/random/{max_value}")
+async def get_random_number(max_value: int, tags=["Random Playground"]):
```
martin-martin, line 73: Consistent with the rest:

```suggestion
@app.get("/random/{max_value}", tags=["Random Playground"])
async def get_random_number(max_value: int):
```

fastapi-python-web-apis/main.py
```diff
@@ -0,0 +1,174 @@
...
+    return {"max": max_value, "random_number": random.randint(1, max_value)}
+
+
+@app.get("/random-between", tags=["Random Playground"])
+def get_random_number_between(
+    min_value: Annotated[
+        int | None,
```
martin-martin, line 80: 🤝 **Handholding:**  You're not doing `int | None` in the tutorial. Does it ever return `None`? Could be nice to have that explained in the tutorial (or removed here, if it's unnecessary).

fastapi-python-web-apis/main.py
```diff
@@ -0,0 +1,174 @@
...
+        Query(
+            title="Minimum Value",
+            description="The minimum random number",
+            ge=1,
+            le=1000,
+        ),
+    ] = 1,
+    max_value: Annotated[
+        int | None,
```
martin-martin, line 89: Same here

fastapi-python-web-apis/main.py
```diff
@@ -0,0 +1,174 @@
...
+        "min": min_value,
+        "max": max_value,
+        "random_number": random.randint(min_value, max_value),
+    }
+
+
+@app.post(
+    "/items", response_model=ItemResponse, tags=["Random Items Management"]
+)
+def add_item(item: Item):
```
martin-martin, line 113: Should all of these be `async def`? Currently, only the first two are, same as you show in the tutorial. But you also mention that readers could update more to async, and that it's important to know which functions to asyncify and which ones not.

So, just wondering if that's intentional or leftover. I guess from my understanding, none of these endpoints benefit from being async (def also not `home()`) :P So just thinking that the code as a "best practice" should likely either commit to showing `async def` for all, or none, to avoid people wondering whether they've missed a subtle detail. Unless I _have_ missed one, of course! :)

PR #702, opened 2025-09-24
Add supporting code for Python Optional Arguments tutorial

martin-martin:
**Where to put new files:**

4. 🎉

### DIFF
--- a/python-optional-arguments/optional_params.py
+++ b/python-optional-arguments/optional_params.py
@@ -13,12 +13,12 @@ Run the module directly to see a short demo.
 
 
 def show_list(shopping_list, include_quantities=True):
-    print()
     for item_name, quantity in shopping_list.items():
         if include_quantities:
             print(f"{quantity}x {item_name}")
         else:
             print(item_name)
+    print()
 
 
 def add_item(item_name, quantity, shopping_list=None):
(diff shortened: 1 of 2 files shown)


martin-martin:
@bzaczynski Copilot suggested to add quantity checks to the two functions that add to the shopping list via `*args` and `**kwargs`. I accepted, I think it makes sense, but decided not to make that same update in the tutorial text.

IMO the tutorial focuses the code on the essential aspect (how do `*args` and `**kwargs` work, with an example) so didactically it makes sense to keep those functions as small as possible. However, I don't mind readers getting the (slightly) more code with the extra check with their materials download and reason about the difference themselves. Seems like a helpful learning experience. LMK if you disagree :)

python-optional-arguments/optional_params.py
```diff
@@ -0,0 +1,86 @@
...
+        if include_quantities:
+            print(f"{quantity}x {item_name}")
+        else:
+            print(item_name)
+
+
+def add_item(item_name, quantity, shopping_list=None):
+    """Add (or increment) an item in a list using the safe 'None' default."""
+    if shopping_list is None:
+        shopping_list = {}
```
bzaczynski, line 27: 🔧 **Technical Recommendation**

The name `shopping_list` may imply Python's `list` data structure, while we actually use the variable to store a Python dictionary. How about renaming this variable to `shopping_items` or `item_quantities` instead?

bzaczynski (approved):
@martin-martin This looks good to me. I'd only consider renaming some of the variables for better clarity, but it's not a big deal at all!

PR #758, opened 2026-04-21
Sample code for: Implementing Interfaces in Python: ABCs and Protocols

lpozo:
Code examples for the [Implementing Interfaces in Python: ABCs and Protocols](https://realpython.com/python-interface/) tutorial.

Generated from the tutorial markdown with the `post-materials` skill, then verified locally before pushing.


bzaczynski (approved):
LGTM!

PR #782, opened 2026-06-14
Clarity Ruff installation before lint commands

asnamobin-hue:
This PR adds a small clarification in the README to explicitly mention that Ruff must be installed before running linting commands.

This helps prevent confusion for new contributors who may try running the commands without installing Ruff first.


asnamobin-hue:
Thanks! I’ve added a small clarification for Ruff installation to help new contributors.

